PR #13, opened 2018-02-21
Update toolbar screenshot.

NumberByColors:


NumberByColors:
@acearun We want the screenshot updating immediately with the pencil icon, correct?

NumberByColors:
Oh, that's a change from several weeks ago. We got rid of all of the icon labels.

NumberByColors:
I think he said to change it to the pencil now, then change it again when the icon change is deployed.

PR #16, opened 2018-03-06
Bring back N-Period Retention to the Usage category

acearun:


Workbooks/Usage/Retention - N Period/settings.json
```diff
@@ -0,0 +1,6 @@
+{
+    "name":"N-Period Retention",
+    "author": "Microsoft",
+    "order": 1200,
+    "visibleInGallery": false
```
ericc1103: So we are not going to display this within Gallery?

acearun: The real version is still under Retention. This one is a hidden template until we fix-up the path in the cards (I broke it when I moved the report)

PR #17, opened 2018-03-08
Users/echong/schema

ericc1103:
Added '@schema' uri which required by localization team so that they can identify it as a localizable file.
Category related localization changes will be the next item to work on.
FYI, the actual schema file doesn't exist and it is not required. It is just used as a placeholder.


gardnerjr:
this is another thing to add to your list of checks that have to pass before merging a PR after this?
all the files with the appropriate extensions shoudl have schema if the author is microsoft?

PR #23, opened 2018-04-06
Add heatmap table coloring to the retention templates

acearun:


Workbooks/Retention/Retention - Bracket/Retention - Bracket.workbook
```diff
@@ -162,7 +162,20 @@
         "size": 0,
         "aggregation": 0,
         "showAnnotations": false,
-        "visualization": "table"
+        "visualization": "table",
+        "gridSettings": {
+          "formatters": [
+            {
+              "columnMatch": "Day",
```
gardnerjr: this one only supports day?  (the one above has week|month|day)

acearun: Brackets are special - they can be any length - so we need to do it by days.

PR #26, opened 2018-05-16
Added the galleries property and assigned default values. visibleInGallery is not removed for the backward compat. Will remove it once the latest version is deployed to AIMON and PROD.

ericc1103:
Added the galleries property and assigned default values. visibleInGallery is not removed for the backward compat. Will remove it once the latest version is deployed to AIMON and PROD.


ericc1103:
FYI, visibleInGallery is not removed for the backward compat. Will remove it once the latest ALMIntegration is deployed to AIMON and PROD.

PR #27, opened 2018-06-01
Users/echong/remove visible in gallery

ericc1103:
- Added TSG templates.
- Removed unused properties.


gardnerjr:
since it is a new one can it have a description set so that gets shown in the gallery UI?

ericc1103:
@gardnerjr Newer version doesn't really care whether it has includeGallery or not.(All of them are visible anyway) It returns all if client doesn't specify type and resourceType.

gardnerjr:
i know the newer version doesn't use it, i'm making sure that newest version is deployed everywhere so the items don't all disappear

ericc1103:
It has been tested in INT env.

gardnerjr:
again, that wasn't the question.  as soon as you deploy this, the field goes away in prod.  so are all the service changes deployed all the way through?

ericc1103:
There is no field is removed other than includeInGallery which doesn't required by a service. Added field won't impact the the existing service.

gardnerjr:
so the answer is "yes, nothing currently deployed to prod references this field anymore" ?

so the prod ALM service doesn't look at this field.
and the prod portal code does not look at this field.

ericc1103:
Yes.

PR #32, opened 2018-06-28
Adding Azure Storage performance templates

acearun:


Workbooks/Storage-Performance/Blob Storage Metrics/settings.json
```diff
@@ -0,0 +1,6 @@
+{
+    "$schema": "https://github.com/Microsoft/Application-Insights-Workbooks/blob/master/schema/settings.json",
+    "name":"Blobs",
+    "author": "Microsoft",
+    "galleries": [{ "type": "performance-storage", "resourceType": "microsoft.insights/components", "order": 200 }]
```
gardnerjr: is this intentionally ai resource type in these?

Workbooks/Storage-Performance/Blob Storage Metrics/settings.json
```diff
@@ -0,0 +1,6 @@
+{
+    "$schema": "https://github.com/Microsoft/Application-Insights-Workbooks/blob/master/schema/settings.json",
+    "name":"Blobs",
```
gardnerjr: should these all have descriptions?

Workbooks/Storage-Performance/Blob Storage Metrics/settings.json
gardnerjr: ugh, but why. its text right here, why not just set them now instead of a PR and fixing them in another PR.  3 steps instead of 1?  :|

PR #34, opened 2018-07-12
Fix storage account template

acearun:


Workbooks/Storage-Performance/Storage Account Metrics/Storage Account Metrics.workbook
```diff
@@ -127,7 +133,7 @@
             }
           }
         ],
-        "resourceType": "microsoft.insights/components"
+        "resourceType": "microsoft.resourcegroup/resourcegroups"
```
vutran01: this won't work until eric check in support for this next week?

PR #36, opened 2018-07-17
Add SQL Database templates

acearun:


Workbooks/SqlDatabase/SQL Database Resource Limits/SQL Database Metrics.workbook
```diff
@@ -0,0 +1,398 @@
...
+      },
```
gardnerjr: this section here "when encountering high session or worker utilization..." is duplicated?

PR #43, opened 2018-08-09
Fix Apdex template to use default time picker

acearun:


Workbooks/Performance/Apdex/Apdex.workbook
```diff
@@ -54,12 +52,91 @@
...
+                  "isInitialTime": false,
+                  "grain": 1,
+                  "useDashboardTimeRange": false
+                }
+              ],
+              "allowCustom": true
+            },
+            "value": {
+              "durationMs": 604800000,
+              "createdTime": "2018-08-09T15:40:35.174Z",
```
gardnerjr: we should investigate if there's a way to make serialized version of these that only have the duration to make all this smaller.  somehow without actually breaking anything else though....

PR #44, opened 2018-08-23
Add Insight templates for App Service and App Insights.

acearun:


Workbooks/App Services - App Metrics/Failure Insights/settings.json
```diff
@@ -0,0 +1,9 @@
+{
+    "$schema": "https://github.com/Microsoft/Application-Insights-Workbooks/blob/master/schema/settings.json",
+    "name":"Failure Insights",
+    "author": "Microsoft",
+    "galleries": [
+        { "type": "failure-websites", "resourceType": "microsoft.resources/subscriptions/resourcegroups", "order": 100 },
```
ericc1103: both of 'order' are 100

acearun: They are in different galleries (one in failures and another in perf). Both supposed to be the first ones shown in their galleries.

PR #47, opened 2018-08-23
Added trivial schema files for cohort and workbook

ericc1103:
Initial version that doesn't contain anything just to avoid 404 error. Will need to provide schema later.


gardnerjr:
would also be interesting if we did make schemas if they could be used to validate json in the PRs themselves?

PR #62, opened 2018-11-13
Added workbook for At Scale Performance - Azure Monitor for VMs

nasundar:
This is an initial attempt to provide a workbook for at-scale performance view of Azure Monitor for VMs.
This contains 
1. Five curated charts to display charts which are available in Azure Monitor. (with top 10 machines)
2. Grid containing top 100 machines based on selected perf counter.
3. Top 5 machines chart based on selected perf counter.

image


gardnerjr:
there's also a `counter` param that's visible but I think you meant to have hidden when not editing?

nasundar:
Fixed ".workbook" extension. Bad typo. :)

**Workspaces dropdown**: Yes I will passing the selected workspace when the workbook is open directly. But I also want to give the flexibility to the user to select a different workspace if needed. 
Also when user opens workbook from gallery, user might want to selected other workspaces.

Can you please elaborate on what you suggested in "adding LA workspace at the top level"?

**Counter**: The counter dropdown needs to be visible, so user can change it to see top 100 machines based on the counter.

nasundar:
Addressing review comments

1. Changed workspaces dropdown to all subscriptions.
2. Fixed aggregation for P90th. 
3. Added to Virtual Machines gallery as well.

PR #64, opened 2018-11-13
Add Workbook for Single VM Performance - Azure Monitor for VMs

andrewki-msft:
Initial attempt at workbook for Single VM performance view of Azure Monitor for VMs. This contains the following:

* Seven charts mirrored from Single VM Insights page
* Grid where user can select a machine to view stats
* Chart where user can view any counter

Initial state
image

Select a workspace and a VM
image

<edit: updated screenshot>

### DIFF
--- a/Workbooks/Virtual Machines/Virtual Machine Performance/Single Virtual Machine Performance.workbook
+++ b/Workbooks/Virtual Machines/Virtual Machine Performance/Single Virtual Machine Performance.workbook
@@ -1,11 +1,11 @@
 {
   "version": "Notebook/1.0",
-  "isLocked": false,
+  "isLocked": true,
   "items": [
     {
       "type": 1,
       "content": {
-        "json": "# Single Virtual Machine Performance\n\n*Click on a row to see trends for only that machine*\n---"
+        "json": "# Single VM Performance Analysis"
       },
       "conditionalVisibility": null
     },
@@ -14,9 +14,7 @@
       "content": {
         "version": "KqlParameterItem/1.0",
         "query": "",
-        "crossComponentResources": [
-          "{Workspace}"
-        ],
+        "crossComponentResources": [],
         "parameters": [
           {
             "id": "8f72709f-725a-41e6-82a2-8d067e38a261",
@@ -24,14 +22,19 @@
             "name": "Workspace",
             "type": 5,
             "isRequired": true,
+            "query": "where type =~ \"Microsoft.OperationalInsights/workspaces\"\r\n| project id",
+            "crossComponentResources": [
+              "value::selected"
+            ],
+            "value": "value::1",
             "isHiddenWhenLocked": false,
             "typeSettings": {
-              "resourceTypeFilter": {
-                "microsoft.operationalinsights/workspaces": true
-              },
-              "additionalResourceOptions": []
+              "additionalResourceOptions": [
+                "value::1"
+              ]
             },
-            "value": "/subscriptions/f9cf50ed-d814-4b95-b76c-c709845f84c9/resourceGroups/mms-eus/providers/Microsoft.OperationalInsights/workspaces/admdemo"
+            "timeContextFromParameter": null,
+            "resourceType": "microsoft.resourcestopology/resources"
           },
           {
             "id": "24eb29a6-309b-4b55-aebd-87d3a2057b3f",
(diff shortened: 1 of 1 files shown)


nasundar:
1. You can probably hide the time grain - it just looks confusing
2. You can set conditionally display charts below since they are based on an action on the grid and not on the top parameters.

gardnerjr:
I would always include time range parameters in workbooks, for a few reasons:
1) it makes it obvious for the user to both see and know what the time range in the charts are

2) if you don't, it might also not be obvious that if you don't explicitly specify time range/parameters in queries, you're only getting the last 24 hours!)

3) pinning support is coming in the next few days, and if you don't use time range parameters, the pinned items will be "stuck" with whatever time range was set when they were pinned, instead of working off of the dashboard's time range

Workbooks/Virtual Machines/Virtual Machine Performance/Single Virtual Machine Performance.workbook
```diff
@@ -0,0 +1,515 @@
+{
+  "version": "Notebook/1.0",
+  "isLocked": false,
```
gardnerjr: for `isLocked` right here. you want it to be true, otherwise when the new workbook is created, it will be in "edit" mode instead of the normal "viewer" mode.

Workbooks/Virtual Machines/Virtual Machine Performance/Single Virtual Machine Performance.workbook
```diff
@@ -0,0 +1,515 @@
...
+            "type": 5,
+            "isRequired": true,
+            "isHiddenWhenLocked": false,
+            "typeSettings": {
+              "resourceTypeFilter": {
+                "microsoft.operationalinsights/workspaces": true
+              },
+              "additionalResourceOptions": []
+            },
+            "value": "/subscriptions/f9cf50ed-d814-4b95-b76c-c709845f84c9/resourceGroups/mms-eus/providers/Microsoft.OperationalInsights/workspaces/admdemo"
```
gardnerjr: in a template, you'll want to remove specific values from parameters.  for resource params, you might want to add "any one" options so that by default this will inherit whatever values are available.   (this template would show a value but not work for people who can't see admdemo!)

Workbooks/Virtual Machines/Virtual Machine Performance/Single Virtual Machine Performance.workbook
```diff
@@ -0,0 +1,515 @@
...
+              "RHEL75"
+            ]
+          },
+          {
+            "id": "d495f8c1-c6b2-490e-8eb6-43d151bc20c0",
+            "version": "KqlParameterItem/1.0",
+            "name": "Counter",
+            "type": 2,
+            "isRequired": true,
+            "query": "Perf\r\n| where TimeGenerated {TimeRange}\r\n| summarize by CounterName, ObjectName, CounterText = strcat(ObjectName, ' > ', CounterName)\r\n| order by ObjectName asc, CounterText asc\r\n| project Counter = pack('counter', CounterName, 'object', ObjectName), CounterText",
```
gardnerjr: does this actaully work?  it looks like this parameter depends on time range but actually comes *before* time range? (time range is the *next* parameter?)

Workbooks/Virtual Machines/Virtual Machine Performance/Single Virtual Machine Performance.workbook
gardnerjr: as a test, sign into some other random azure account (like a personal one that doesn't have access to admdemo, and paste your template into advanced mode and see if it works.  that's how I test most people's templates.  @ericc1103 we need a way in the portal to point workbooks at a pull request somehow and have it show whatever templates are in that PR?

ericc1103: @gardnerjr Good suggestion. Could you add it into the backlog since you have the list?

andrewki-msft: Thank you for the comments and detailed instructions on how to test, I have tested it on a separate account with a different subscription.

andrewki-msft:
If I'm navigating to this workbook from a VM in the same journey, what variables should I reference within the Workbook itself in order to carry over the workspace and VM?

nasundar:
Please checkin after @gardnerjr or @ericc1103 approves this PR.

andrewki-msft:
Okay, going ahead with merge.

gardnerjr:
@nasundar  that is correct.  if the parameter was a multi select dropdown/resource, then it should be an **array** with that single value inside, but yes.

PR #67, opened 2018-11-20
Fix Single VM chart workbook

andrewki-msft:
Change the `computer` parameter to sentence case (i.e. `Computer`) so it
will correctly receive the current computer name when the user navigates
to this workbook from a single VM chart.

- Change `computer` hidden parameter to `Computer`

- Adjust kusto query to remove unused clause

image

### DIFF
--- a/Workbooks/Virtual Machines/Virtual Machine Performance/Single Virtual Machine Performance.workbook
+++ b/Workbooks/Virtual Machines/Virtual Machine Performance/Single Virtual Machine Performance.workbook
@@ -220,10 +220,9 @@
           {
             "id": "6e1380f6-5694-4457-8958-7a4a6dcb7d56",
             "version": "KqlParameterItem/1.0",
-            "name": "computer",
+            "name": "Computer",
             "type": 1,
             "isRequired": false,
-            "value": "",
             "isHiddenWhenLocked": true,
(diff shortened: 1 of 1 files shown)


Workbooks/Virtual Machines/Virtual Machine Performance/Single Virtual Machine Performance.workbook
```diff
@@ -337,7 +336,7 @@
       "type": 3,
       "content": {
         "version": "KqlItem/1.0",
-        "query": "let metric = dynamic({Counter});\r\nPerf\r\n| where TimeGenerated {TimeRange}\r\n| where Computer in ({VirtualMachines}) or '*' in ({VirtualMachines})\r\n| where Computer == '{computer}' or 'All' == '{computer}'\r\n| where ObjectName == metric.object and CounterName == metric.counter\r\n| summarize {Metrics} by bin(TimeGenerated, totimespan('{grain}'))",
```
vutran01: are you removing the All option?

andrewki-msft: Yes, that clause is not necessary since only a single VM can be displayed at a time

PR #70, opened 2018-11-26
Move performance Analysis workbooks to virtual machines directory

nasundar:
1. Move 'Performance Analysis' and 'Network Dependencies' workbooks to Virtual Machine Directory.
2. Change Workspaces dropdown to 'Workbook Resource Param' type instead of 'Resource Graph query'.


Workbooks/Virtual Machines - Performance Analysis/Performance Analysis for a group of VMs/Performance Analysis for a group of VMs.workbook
```diff
@@ -15,7 +15,7 @@
         "version": "KqlParameterItem/1.0",
         "query": "",
         "crossComponentResources": [
-          "{Workspaces}"
+          "value::all"
```
gardnerjr: are you sure you want all?  do you want "any one" or whatever?

nasundar:
Also, I’m seeing this for whatever workspaces I can see:

 
image


You might need to look at tableifexists or columnifexists or whatever to harden these?

nasundar:
image

Workbooks/Virtual Machines - Performance Analysis/Performance Analysis for a group of VMs/Performance Analysis for a group of VMs.workbook
nasundar: Thank you. I had set it as 'Any One' only, not sure why it picked up 'All'. 
I went back to setting for the param and reset it - apparently it removed everything inside "crossComponentResources".

I think this should work.

PR #72, opened 2018-11-27
Enabled CI build and added json validation logic

ericc1103:


test/jsonValidationTest.js
```diff
@@ -0,0 +1,124 @@
...
+                done();
+        });
+    });
+
+    it('Verifying workbook settings.json files', function(done){
+        browseDirectory(cohortPath, (error, results) => {
+            if (error) throw error;
+            results.filter(file => file.substr(-13) === 'settings.json')
+                .forEach(file => {
+                    console.log(file);
```
gardnerjr: log needed?

test/jsonValidationTest.js
```diff
@@ -0,0 +1,124 @@
...
+                    validateJsonString(file);
+                });
+            
+                done();
+        });
+    });
+
+    it('Verifying workbook category json files', function(done){
```
gardnerjr: for category and settings, do you want further validation?
they have the right fields, etc.  right now this just validates that they are valid json.  so I could have a categoryResources.json that's literally `[]` and that would be valid json.

test/jsonValidationTest.js
ericc1103: Won't hurt. :-)

test/jsonValidationTest.js
ericc1103: That would be the next step. We have a lot to add.

PR #73, opened 2018-11-27
Added more detail validation for settings and category

ericc1103:


test/jsonValidationTest.js
```diff
@@ -5,120 +5,167 @@ const fs = require('fs');
...
 
-    it('Verifying .cohort files', function(done){
+    it('Verifying .cohort files', function (done) {
         let failedList = [];
         browseDirectory(cohortPath, (error, results) => {
             results.filter(file => file.substr(-7) === '.cohort')
-                .forEach(file => {                    
-                    validateJsonString(file)
+                .forEach(file => {
+                    validateJsonStringAndGetObject(file)
```
gardnerjr: I *think* you can go all "javascript" on this block, and just
```
.forEach(validateJsonStringAndGetObject);
```
if you're just passing the param directly to the method

test/jsonValidationTest.js
```diff
@@ -5,120 +5,167 @@ const fs = require('fs');
...
     });
 
-    it('Verifying cohort category files', function(done){
+    it('Verifying cohort category files', function (done) {
         browseDirectory(cohortPath, (error, results) => {
             if (error) throw error;
             results.filter(file => file.substr(-22) === 'categoryResources.json')
                 .forEach(file => {
-                    validateJsonString(file);
+                    let json = validateJsonStringAndGetObject(file);
```
gardnerjr: nit: it isn't "json" after this call, it's an actual object.

test/jsonValidationTest.js
ericc1103: @gardnerjr I am not sure what you mean by that.. Could you elaborate more?

test/jsonValidationTest.js
```diff
@@ -5,120 +5,167 @@ const fs = require('fs');
...
+function validateJsonStringAndGetObject(file) {
+    let json = fs.readFileSync(file, 'utf8');
+    let obj = TryParseJson(json)
+    assert.isTrue(obj != null, 'Invalid json format with \'' + file + '\'');
+    return obj;
+}
 
+function validateSettingsForCohort(settings) {
+    if (!settings.hasOwnProperty('$schema')) {
+        assert.fail("The $schema field is missing with '" + file + "'");
```
gardnerjr: this block is repeated like 100 times?

```
function validateSetting(obj: any, name: string, file: string) {
    if (!obj.hasOwnProperty(name)) {
        assert.fail( file + ":  property " + name + " is missing" );
    }
}
```

then use it like:
```
["name", "$schema", "author"].forEach( field => validateSetting(settings, field, file) );
```

in each test you just change what named fields are validated

test/jsonValidationTest.js
ericc1103: Done

test/jsonValidationTest.js
```diff
@@ -5,120 +5,167 @@ const fs = require('fs');
...
-      })();
+        })();
     });
-  };
+};
 
+function validateJsonStringAndGetObject(file) {
+    let json = fs.readFileSync(file, 'utf8');
+    let obj = TryParseJson(json)
+    assert.isTrue(obj != null, 'Invalid json format with \'' + file + '\'');
```
gardnerjr: why not catch the error and actually output it in the assert?  if the file is invalid, why not explain why right here?

test/jsonValidationTest.js
gardnerjr: i believe you can replace this block:

```.forEach(file => { 
validateJsonStringAndGetObject(file);
});
```

with `.forEach(validateJsonStringAndGetObject);`

test/jsonValidationTest.js
ericc1103: Good suggestion

test/jsonValidationTest.js
ericc1103: I will just leave as is since it is not significant and leave it for more readability. :-)

PR #74, opened 2018-11-29
Added usage type

ericc1103:


Workbooks/Usage/Retention - N Period/settings.json
```diff
@@ -1,6 +1,17 @@
 {
```
acearun: Is this template visible? Should not be.

PR #75, opened 2018-12-05
Add crossComponentResources to Single VM workbook

andrewki-msft:
I added `crossComponentResources` or the workbook will not correctly read in the passed in resources (workspace and computer ID). Also removed blank items.


andrewki-msft:
The template wasn't the issue so no need to update, closing this PR

PR #87, opened 2018-12-12
First draft of Security and Audit workbook

andrewki-msft:
Port over various charts from OMS Security and Audit panel


nasundar:
1. You can probably set a default message if no records are available like 'No Potential Malicious Traffic.'
image
2. Increase the number of rows from default 250 to more.
image
3. May be give out a time range at the top, The queries doe not have any time bounds. They will be potentially returning millions of records for an active workspace.
4. May be make some headers a little bold (#####).
image

PR #96, opened 2019-01-15
Fix Single VM Perf Analysis workbook

andrewki-msft:
Update description text and select a default counter so the workbook
doesn't show a broken view when opened from gallery.

image

### DIFF
--- a/Workbooks/Virtual Machines - Performance Analysis/Performance Analysis for a Single VM/Performance Analysis for a Single VM.workbook
+++ b/Workbooks/Virtual Machines - Performance Analysis/Performance Analysis for a Single VM/Performance Analysis for a Single VM.workbook
@@ -5,122 +5,7 @@
     {
       "type": 1,
       "content": {
-        "json": "# Single VM Performance Analysis\nAnalyze the performance counters (processor, memory, disk, etc.) of a single VM. You can also customize this workbook by clicking on the `Edit` button in the toolbar."
-      }
-    },
-    {
-      "type": 9,
-      "content": {
-        "version": "KqlParameterItem/1.0",
-        "query": "",
-        "crossComponentResources": [],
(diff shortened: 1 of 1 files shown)


nasundar:
In all the queries I still see "| where Computer == Computer". Since you are using resource centeric queries, you dont need to add this condition.

Workbooks/Virtual Machines - Performance Analysis/Performance Analysis for a Single VM/Performance Analysis for a Single VM.workbook
```diff
@@ -159,7 +163,8 @@
             "query": "Perf\r\n| where TimeGenerated {TimeRange}\r\n| summarize by CounterName, ObjectName, CounterText = strcat(ObjectName, ' > ', CounterName)\r\n| order by ObjectName asc, CounterText asc\r\n| project Counter = pack('counter', CounterName, 'object', ObjectName), CounterText",
             "isHiddenWhenLocked": false,
             "timeContextFromParameter": null,
-            "resourceType": "microsoft.compute/virtualmachines"
+            "resourceType": "microsoft.compute/virtualmachines",
+            "value": "{\"counter\":\"% Used Inodes\",\"object\":\"Logical Disk\"}"
```
nasundar: If there is no data to display, this will display an empty chart. Can you please add a No data message and display that when there is no data instead of empty chart.

Something like this but with appropriate computer name:
image


PS: Also check if this is one of the important counters, we can probably use a more widely used counter.

nasundar:
We have using different names for parameter dropdowns in different workbooks. Can we have all of those use same name?

I did not see this our meeting notes containing all the changes discussed. Can you please add this as well. It need not be in this PR.

PR #104, opened 2019-01-28
Update Single VM workbook

andrewki-msft:
Further polish Single VM performance analysis workbook to bring it more
in line with its AtScale counterpart. Also delist a duplicate workbook
since that has been incorporated into this workbook. Update cosmetic
look-and-feel and add helper text to aid end-user in navigating around
this workbook.

- Rename workbook to "Performance" in Single VM gallery

- Delist "Virtual Machine Performance" workbook from Single VM gallery

- Add and update helper text

- Cosmetic changes for easier readability

- Align parameter names with AtScale Performance workbook

- Minor fix in last chart to use correct parameter

image


nasundar:
I cannot validate this, is there any known issue with Workbooks or incident going on with Draft?

image

PR #111, opened 2019-02-07
SignIn Usage Workbook Update and Addition of SignIn Failure Analysis, Conditional Access Workbooks.

KranthiKiranerusu:


Workbooks/Azure Active Directory TroubleShooting/categoryResources.json
```diff
@@ -0,0 +1,3 @@
+{
+    "en-us": {"name":"TroubleShoot", "description": "Provides insights on Troubleshooting", "order": 200}
```
acearun: We made a recent change that allows you to specify categories within the gallery section of the template settings.json. That way your categories are not constrained to the folder structure of github - particularly interesting if the template needs to show up in different galleries under unique categories. However, it not strictly required for your scenario here. If you are interested in find out out, please reach out to Eric Chong.

KranthiKiranerusu: That is indeed useful. Will work with Eric Chong on that

PR #112, opened 2019-02-07
Hotfix Connections Overview

andrewki-msft:
Fixed computers not showing in the table list for AtScale Connections
Overview. Recent changes in default workbooks query broke this workbook
causing the computer name to be wrongly populated.

Preview
image

### DIFF
--- a/Workbooks/Virtual Machines - Network Dependencies/Connections Overview/Connections Overview.workbook
+++ b/Workbooks/Virtual Machines - Network Dependencies/Connections Overview/Connections Overview.workbook
@@ -52,8 +52,8 @@
             "version": "KqlParameterItem/1.0",
             "name": "ComputerName",
             "type": 1,
-            "isRequired": true,
-            "query": "print '{Computer:label}'",
+            "isRequired": false,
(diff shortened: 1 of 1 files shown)


Workbooks/Virtual Machines - Network Dependencies/Connections Overview/Connections Overview.workbook
```diff
@@ -52,8 +52,8 @@
             "version": "KqlParameterItem/1.0",
             "name": "ComputerName",
             "type": 1,
-            "isRequired": true,
-            "query": "print '{Computer:label}'",
+            "isRequired": false,
+            "query": "print 'undefined'",
```
nasundar: I thought you were planning on creating a different workbook for single VM. Is that not the case?

andrewki-msft: Yes, a separate workbook for Single VM exists, it's just that the cleanup in this current workbook hasn't been fully completed. Will do this in a future commit. Thanks.

PR #118, opened 2019-02-13
Update Connections Overview for Single VM

andrewki-msft:
Bring Connections Overview workbook for Single VM to parity with its
AtScale counterpart. Includes all fixes such as inbound connections and
additional filters.

Preview:
image

### DIFF
--- a/Workbooks/Virtual Machines - Network Dependencies/Connections Overview - Single VM/Connections Overview.workbook
+++ b/Workbooks/Virtual Machines - Network Dependencies/Connections Overview - Single VM/Connections Overview.workbook
@@ -3,32 +3,23 @@
   "isLocked": true,
   "items": [
     {
+      "type": 1,
+      "content": {
+        "json": "<h1>Connections Overview</h1>"
+      },
+      "conditionalVisibility": null
+    },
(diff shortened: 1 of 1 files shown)


Workbooks/Virtual Machines - Network Dependencies/Connections Overview - Single VM/Connections Overview.workbook
```diff
@@ -2,34 +2,25 @@
...
-              "additionalResourceOptions": [
-                "value::1"
-              ]
-            },
-            "timeContextFromParameter": null
-          },
-          {
-            "id": "67b8a28c-f7f9-4903-883d-4a308a6ffe71",
-            "version": "KqlParameterItem/1.0",
             "name": "Computer",
```
vutran01: where is the Computer param shown in the mockup?

andrewki-msft: The `Computer` param is hidden all the way at the top, it is a dropdown used to select the "default" Computer resource that is passed into this workbook
image

Issue #89, opened 2018-12-13
Update JSON validation test to handle UTF-8 Byte-Order Mark (BOM) error


andrewki-msft:
The JSON validation test fails if Byte-Order Mark (BOM) is present in the file and throws an ambiguous error. The validation test should be updated to warn the user to take corrective action to remove BOM. It is recommended that BOM should not be used in UTF-8.

https://tools.ietf.org/html/rfc3629#page-6

Steps to reproduce:
1. Use notepad++ or some text editor that prepends a UTF-8 BOM
2. Save a new workbook as UTF-8
3. Execute `jsonValidationTest.js`
4. Test will fail, throwing an error saying unexpected "blank" token was found
```
SyntaxError: Unexpected token ﻿ in JSON at position 0
      at TryParseJson (test/jsonValidationTest.js:149:16)
      at validateJsonStringAndGetObject (test/jsonValidationTest.js:115:15)
      at results.filter.forEach.file (test/jsonValidationTest.js:55:21)
      at Array.forEach (<anonymous>)
      at browseDirectory (test/jsonValidationTest.js:54:18)
      at next (test/jsonValidationTest.js:96:31)
      at /home/travis/build/Microsoft/Application-Insights-Workbooks/test/jsonValidationTest.js:102:25
      at next (test/jsonValidationTest.js:96:31)
      at /home/travis/build/Microsoft/Application-Insights-Workbooks/test/jsonValidationTest.js:106:21
      at FSReqCallback.oncomplete (fs.js:161:5)
```

Fix diff for the workbook to remove the BOM:
```
@@ -1,4 +1,4 @@                                                                                                                                                       
-<EF><BB><BF>{                                                                                                                                                        
+{                                                                                                                                                                    
   "version": "Notebook/1.0",
```

gardnerjr:
I'm tryng to add @ericc1103 to who it is assigned to and somehow it autocompleted to 04diiguyi?  wth github?

ericc1103:
That is not me... Now sure why..


Thanks,

Eric

________________________________

gardnerjr:
from an ancient archive of nodejs 0.x:

https://github.com/nodejs/node-v0.x-archive/issues/1918

PR #125, opened 2019-02-21
Users/echong/readme

ericc1103:


README.md
```diff
@@ -161,3 +161,52 @@ There are three ways of creating a template.
     ![Image of default template](https://github.com/Microsoft/Application-Insights-Workbooks/blob/master/_assets/SavedList.png)
 
 6. Modify report as you wish and click "Advanced Editor" button from the menu. Copy all contents and create a file like "your custom template name.template". Please make sure file name ends with '.workbook'.
+
+## How to associate any existing template to new category in Workbooks
```
gardnerjr: should this be "Additional categories" ?  this allows you to put workbooks in a whole bunch of places, right, not just 2?

Workbooks/categoryResources.json
```diff
@@ -0,0 +1,13 @@
+{
+    "categories": [{
+        "key": "YourUniqueCategoryKey",
```
gardnerjr: instead of placeholder stuff, should we use this to create a new category like "Samples" or something, and put some sample workbooks into that?

there's another story somewhere to allow feature flags (or extra ui) to include extra categories/galleries.  this might be a good place to start, with samples?

test/jsonValidationTest.js
```diff
@@ -45,11 +45,11 @@ describe('Validating Cohorts...', () => {
 });
 
 describe('Validating Workbooks...', () => {
```
gardnerjr: while you are touching this do you want to fix #89 ?  it's seems to be a 1 line fix in here.

README.md
```diff
@@ -161,3 +161,52 @@ There are three ways of creating a template.
...
+We are now supporting associating any exiting templates to new category (virtual category). Previously, a category is always associated with templates by a folder structure but this requires to create physical folder structure which requires copying existing templates. This would introduce a lot of maintain overhead of updating duplicated templates.
```
gardnerjr: was

gardnerjr: maintainance

README.md
```diff
@@ -161,3 +161,52 @@ There are three ways of creating a template.
...
+        "author": "Microsoft",
+        "galleries": [
+            {
+                "type": "workbook",
+                "resourceType": "microsoft.insights/components",
+                "order": 400
+            },
+            {
+                "type": "workbook",
+                "resourceType": "Azure Monitor",
```
gardnerjr: nothing indicates that this second item is the new part, until you read the note below.  it would be nice if json had comments.  so you might want to mention this before the code instead of after?

it might also be worth splitting up the readme into multiple markdown files, with the main one having contributing instructions and links to the other docs?

README.md
gardnerjr: (GitHub doesn't seem to do a good job of marking what was highlighted when you add a comment. :(  "maintain" in "this would introduce a lot of maintain" should be "maintenance")

gardnerjr: (GitHub doesn't seem to do a good job of marking what was highlighted when you add a comment. :(  "is" in "previously, a category is always" should be "was").

or you could explain that the primary way things work is via folder structure, and that you can *additionally* do it through virtual categories.

test/jsonValidationTest.js
ericc1103: I will handle it separately. Not sure there is an easy way to detect BOM.

gardnerjr: as I commented in that thing, it's a one line change:

> from an ancient archive of nodejs 0.x:
> nodejs/node-v0.x-archive#1918
> it looks like the fix is to just strip it out after reading from the file:
> `body = body.replace(/^\uFEFF/, '');`

Workbooks/categoryResources.json
gardnerjr: you marked this resolved but i don't see any comment or change?   why does this *real* categoryResources.json have placeholder example content?  shouldn't this have *real* content?  its ok if the docs have placeholder values like "YourUniqueCategoryKey" but why does this real file?

ericc1103: We don't have any virtual category yet until Scott's team adds one. Once they add one, I will remove it.

test/jsonValidationTest.js
ericc1103: To unblock Scott's team I will deal with it separately since it may require more testing and investigation as we discussed.

PR #132, opened 2019-03-05
Workbook for Intune Audity Activity

kumarvaranasi:


Workbooks/Intune Audit/Audit/IntuneAuditActivity.workbook
```diff
@@ -0,0 +1,390 @@
+{
+  "version": "Notebook/1.0",
+  "isLocked": false,
```
gardnerjr: for sure you want this to be `"isLocked": true,` (or just remove the line entirely)

gardnerjr:
tagged @ericc1103  and @acearun to validate the category/folder part.

Workbooks/Intune Audit/Audit/IntuneAuditActivity.workbook
```diff
@@ -0,0 +1,390 @@
...
+        "json": "## Audit Events Trend"
+      },
+      "conditionalVisibility": null
+    },
+    
+    {
+      "type": 3,
+      "content": {
+        "version": "KqlItem/1.0",
+        "query": "IntuneAuditLogs\r\n| summarize count() by bin(TimeGenerated, time(1h))",
```
gardnerjr: for the `bin` part, you probably want this to be something like `bin(TimeGenerated, {TimeRange:grain})` to inherit a default grain from the selected `TimeRange` parameter instead of always being 1 hour?  (from math I can't tell what the time ranges are, but using `{TimeRange:grain}` formatter will get you a reasonable bin grain from a time setting

gardnerjr:
my only other general comment would be: is there a screenshot or a place where we can try it out?  msot of the time just looking at the json isn't exactly obvious what's going on, so pictures are worth a billion words?

Workbooks/Intune Audit/Audit/settings.json
```diff
@@ -0,0 +1,6 @@
+{
+    "$schema": "https://github.com/Microsoft/Application-Insights-Workbooks/blob/master/schema/settings.json",
+    "name":"IntuneAuditActivity",
```
ericc1103: You would probably want to name it like `Intune Audit Activity` since it will be appeared in a gallery as well as it will be localized.

Workbooks/Intune Audit/Audit/IntuneAuditActivity.workbook
```diff
@@ -0,0 +1,390 @@
...
+        "json": "## Audit Events Trend"
+      },
+      "conditionalVisibility": null
+    },
+    
+    {
+      "type": 3,
+      "content": {
+        "version": "KqlItem/1.0",
+        "query": "IntuneAuditLogs\r\n| summarize count() by bin(TimeGenerated, {TimeRange:grain})",
```
acearun: In general, if you have to bin by time, then make-series is a better alternative because it handles buckets with 0 items better. Summarize can cause chart to time-shift if the 0 items are at the beginning or end.

Workbooks/Intune Audit/Audit/IntuneAuditActivity.workbook
kumarvaranasi: Thank you for the feedback. I updated the review with it.

Workbooks/Intune Audit/Audit/IntuneAuditActivity.workbook
kumarvaranasi: Thank you Arun. I will play with make-series and follow up with another checkin.

kumarvaranasi:
Thank you John, Arun and Eric for the feedback.

PR #142, opened 2019-03-12
Private/aruna/fixexistingtemplates

acearun:
Active Users:
image

Apdex:
image

### DIFF
--- a/Workbooks/Performance/Apdex/Apdex.workbook
+++ b/Workbooks/Performance/Apdex/Apdex.workbook
@@ -206,9 +206,49 @@
       "conditionalVisibility": null
     },
     {
-      "type": 1,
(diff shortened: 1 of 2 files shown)


Workbooks/Performance/Apdex/Apdex.workbook
```diff
@@ -191,85 +151,184 @@
           {
             "id": "2f57173e-c1de-4b51-90a6-fc4baee9fec6",
             "version": "KqlParameterItem/1.0",
-            "name": "SegmentBy",
+            "name": "SplitBy",
```
gardnerjr: does users blade pass this in when you open apdex from there?  if so, it won't be hooked up anymore.

PR #143, opened 2019-03-13
Add VMBoundPort workbooks

andrewki-msft:
Add VMBoundPort workbooks for VM Insights AtScale workbooks gallery

Active Ports
image

Open Ports
image


Workbooks/Virtual Machines - Network Dependencies/Open Ports/settings.json
```diff
@@ -0,0 +1,8 @@
+{
+    "$schema": "https://github.com/Microsoft/Application-Insights-Workbooks/blob/master/schema/settings.json",
+    "name": "Open Ports",
+    "author": "Microsoft",
+    "galleries": [
+        { "type": "workbook", "resourceType": "microsoft.operationalinsights/workspaces", "order": 200 }
```
acearun: This is the same order as the other template. Change it up or down based on priority.

Workbooks/Virtual Machines - Network Dependencies/Active Ports/settings.json
```diff
@@ -0,0 +1,9 @@
+{
+    "$schema": "https://github.com/Microsoft/Application-Insights-Workbooks/blob/master/schema/settings.json",
+    "name":"Active Ports",
```
acearun: I can't tell the difference between Active Ports and Open Ports. Maybe a better name would help.

acearun:
In the Open Ports report, surely the computers don't have thousands of ports open? Do you have the wrong aggregation in the query?

acearun:
In the Open Ports report, you have the option of showing a hierarchy of Process > Open Ports with some metric (like bytes sent). The flat list is a bit bland.

Workbooks/Virtual Machines - Network Dependencies/Open Ports/settings.json
andrewki-msft: Updated order of workbooks
image

Workbooks/Virtual Machines - Network Dependencies/Active Ports/settings.json
andrewki-msft: Those were the names that our PM created them with, unfortunately changing the titles would incur changes elsewhere (e.g. blog, etc.) so will leave this be for now.

andrewki-msft:
> In the Open Ports report, surely the computers don't have thousands of ports open? Do you have the wrong aggregation in the query?

Yes, you're right, will fix this

PR #145, opened 2019-03-14
Open workbooks in read-only mode

KranthiKiranerusu:


KranthiKiranerusu:
Updates to Azure Active Directory Workbooks.

PR #158, opened 2019-03-25
Documentation for tiles

acearun:


gardnerjr:
this reminds me that we should add a task for some kind of auto conversion if there are table settings to try to inherit those in the tiles (or possibly graph?) so that if you configure one, the settings tries to inherit that if you switch to another renderer.

PR #176, opened 2019-04-30
move text step headings into chart headings, set no rows text

gardnerjr:
for Bug 1441498: [Screen Reader - Application Insights E2E - Failure Analysis] When focus received 'Failed Operations' and 'All Operations' graph chart, screen reader not given relevant information

switched text steps for chart headers to actual chart headers, added failure text:
image


gardnerjr:
@acearun is it intentional that this report only uses requests as the source of operations?  on apps like insightsportal-prod (which is effectively a client only SPA), there are no requests,  so the whole workbook is always empty?

PR #177, opened 2019-05-01
Changing the template

shijatsu:
to the PM mockup

### DIFF
--- a/Workbooks/Storage/Overview/Overview.workbook
+++ b/Workbooks/Storage/Overview/Overview.workbook
@@ -4,7 +4,7 @@
     {
       "type": 1,
       "content": {
-        "json": "## Storage Account Insights"
+        "json": "# Storage Insights"
       },
       "conditionalVisibility": null,
       "name": "text - 0"
@@ -15,7 +15,7 @@
         "version": "KqlParameterItem/1.0",
         "query": "",
         "crossComponentResources": [
-          "value::selected"
+          "{Subscription}"
         ],
         "parameters": [
           {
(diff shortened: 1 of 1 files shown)


Workbooks/Storage/Overview/Overview.workbook
```diff
@@ -15,7 +15,7 @@
         "version": "KqlParameterItem/1.0",
         "query": "",
         "crossComponentResources": [
-          "value::selected"
+          "{Subscription}"
```
gardnerjr: I don't think this part is right?  this says this very first parameter depends on another parameter called {subscription} that doesn't exist yet?

Workbooks/Storage/Overview/Overview.workbook
```diff
@@ -44,16 +44,19 @@
...
             "typeSettings": {
-              "limitSelectTo": 10,
+              "limitSelectTo": 25,
               "additionalResourceOptions": []
             },
             "timeContextFromParameter": null,
             "queryType": 1,
-            "resourceType": "microsoft.resourcegraph/resources"
+            "resourceType": "microsoft.resourcegraph/resources",
+            "value": [
```
gardnerjr: this actually has a selected subscription inside, should be value::selected or all or whatever, no specific resources should be selected

Workbooks/Storage/Overview/Overview.workbook
```diff
@@ -174,19 +177,82 @@
...
+          },
+          "showBorder": false
+        }
+      },
+      "conditionalVisibility": null,
+      "customWidth": "20",
+      "name": "query - 20"
+    },
+    {
+      "type": 3,
```
gardnerjr: if this health stuff isn't ready, just remove it?  (right now it's still running)

Workbooks/Storage/Overview/Overview.workbook
shijatsu: Fixed in new PR

Workbooks/Storage/Overview/Overview.workbook
shijatsu: Fixed in new PR

Workbooks/Storage/Overview/Overview.workbook
shijatsu: fixed in new PR

shijatsu:
Opened a new PR

PR #178, opened 2019-05-01
Removed health section and made resource name conditionally visible

shijatsu:


Workbooks/CosmosDb/Overview/Overview.workbook
```diff
@@ -416,7 +378,11 @@
       "content": {
         "json": "### Cosmos DB account details: <span style=\"color:blue\"> _{ResourceId:name}_ </span> "
```
gardnerjr: same comment as the other PR, might want to try this on dark theme and see if there's a better way

shijatsu: Removed HTML color code in markup

PR #179, opened 2019-05-01
Changed Grid, resource count

shijatsu:
Metric Grid has a new  version
added resource count tile 
added conditional visibility to resource name

### DIFF
--- a/Workbooks/Storage/Overview/Overview.workbook
+++ b/Workbooks/Storage/Overview/Overview.workbook
@@ -182,11 +182,53 @@
       "name": "parameters - 1"
     },
     {
+      "type": 3,
+      "content": {
+        "version": "KqlItem/1.0",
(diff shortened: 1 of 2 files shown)


Workbooks/Storage/Overview/Overview.workbook
```diff
@@ -451,10 +454,10 @@
       "conditionalVisibility": null,
       "name": "text - 4"
     },
-    {
+       {
       "type": 1,
       "content": {
-        "json": "## Information about storage account _{ResourceId:name}_"
+        "json": "### Storage account details: <span style=\"color:blue\"> _{ResourceId:name}_ </span> "
```
gardnerjr: why is it blue?  on dark theme it will probably be mostly invisible, blue on black background?

Issue #124, opened 2019-02-21
update JSON validation to warn/fail about other various uncommon things


gardnerjr:
1) warning: edit mode - Workbooks that are unlocked (in edit mode) are rare, but possible.  but it's almost always never what you want.

so the validator should warn if a workbook is in edit mode, so we catch it in PRs here.

2) error: `defaultResourceIds`, `fallbackResourceIds` - these should always be not set, they're for workbooks, not for templates

3) warning: `crossComponentResources` that aren't using parameters?

gardnerjr:
edit mode was removed from being in the default output, so (1) above is basically taken care of that way.

#181 would mostly handle all the other ones, by just verifying that the templates don't contain anything with `/subscriptions/[guid]` in them _anywhere_.

PR #181, opened 2019-05-06
attempting resource id validation

gardnerjr:
this pr should fail to validate because there's a workbook that is failing this validation.

so this proves that the test works, then I can fix the workbook in a different pr, then this should work to validate.

### DIFF
--- a/test/jsonValidationTest.js
+++ b/test/jsonValidationTest.js
@@ -52,7 +52,7 @@ describe('Validating Workbooks...', () => {
         browseDirectory(workbookPath, (error, results) => {
             results.filter(file => file.substr(-9) === '.workbook')
                 .forEach(file => {
-                    validateJsonStringAndGetObject(file);
+                    let settings = validateJsonStringAndGetObject(file);
                     validateNoResourceIds(settings, file);
                 });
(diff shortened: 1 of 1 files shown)


gardnerjr:
@ericc1103 why didn't this PR run the validation checks? because no json/etc were changed?

gardnerjr:
actually, it looks like the build/validation part isn't running at all, I don't see build/test output on any of the recent PRs

gardnerjr:
would implement part of #124 for resource ids at least

ericc1103:
@gardnerjr  Yours needs to be checked-in in order to be used.

gardnerjr:
@ericc1103 if you have to complete the PR to run the tests, then anyone could actively prevent everyone from completing any PRs by simply creating a test that always fails and completing the PR.  If the tests are modified as part of a PR the PR should run the tests, no?

ericc1103:
You would have to test your PR by running it locally. I won't run in GitHub until it checked in.

gardnerjr:
completing, but there's the possibility this breaks things things the changes to the tests don't get run as part of this PR...

PR #186, opened 2019-05-13
Fix Performance Workbook in Context-Free Mode

andrewki-msft:
Reduce the number of ARG parameters to prevent 429 errors by utilizing JSON parameters and combining ARG parameters. Also remove comments from all of the queries. Also fix edge case where a user doesn't have a subscription with any log analytics workspaces.

Number of hidden ARG parameters reduced to solve the 429 issue
image

Utilize ordered ARG query to limit randomness
image

Updated help text
image

Error mode
image


andrewki-msft:
Waiting on JSON fix to be deployed to PROD, then can merge this in, but I'm still seeing some occasional problems on PPE with the workbook initialization but at least I am no longer encountering 429 errors.

Workbooks/Virtual Machines - Performance Analysis/Performance Analysis for a Group of VMs/Performance Analysis for a Group of VMs.workbook
```diff
@@ -285,7 +253,7 @@
             "name": "Test",
             "type": 1,
             "isRequired": false,
-            "query": "// {DefaultWorkspace} {ContextFree} {ContextSelection} {RandomSelection} {Selection} {Subscriptions} {Workspaces}\r\nPerf\r\n| take 1",
+            "query": "Perf\r\n| take 1",
```
gardnerjr: these queries don't seem to use time range, either in the query or in the time range picker, is that ok?  this query will be limited to last 24h, is that ok?

andrewki-msft: I think for `Test` parameter we don't want to limit it by any time range but search the entire `Perf` table for any entries (or if the table even exists, at which point the parameter will be empty since it's marked as not required). I didn't understand the second part, are you saying that if I don't include a time range that it will default to the last 24 hours? Because that is not okay since that may give the wrong prompt to the user. I set the Time Range to `Set in query` so my assumption was that it would search the entire table regardless of time range.

image

gardnerjr: if it is set to "set in query", and the query doesn't *explicitly* set a range, then yes, it defaults to 24h.  otherwise, that query could be looking at YEARS of data.  this is the same behavior as the log analytics portal's query stuff.  if you don't specify a time range, you get 24h.  it's always best to set a time range either explicitly in the query text or by picking the parameter.

andrewki-msft: Thank you @gardnerjr, I have updated the `Test` parameter and the corresponding help text to ask the user to now either adjust the time range or select a different workspace, along with onboarding instructions. Also I've added an error text because I still see cases where the `Subscription` parameter fails to initialize so I ask the user to take corrective action as well.

Workbooks/Virtual Machines - Performance Analysis/Performance Analysis for a Group of VMs/Performance Analysis for a Group of VMs.workbook
```diff
@@ -284,47 +240,58 @@
...
+        "resourceType": "microsoft.insights/components"
       },
-      "conditionalVisibility": null
+      "name": "parameters - 2"
     },
     {
       "type": 1,
       "content": {
-        "json": "⚠ This workbook requires [Azure Monitor for VMs (preview)](https://docs.microsoft.com/en-us/azure/azure-monitor/insights/vminsights-overview) enabled in order to show accurate and complete information for your VMs. Follow the link to onboard machines or select a different workspace."
+        "json": "🛑 There was an error initializing this workbook, select an item under the `Subscriptions` dropdown or refresh the workbook."
```
gardnerjr: nit: I personally wouldn't callout "there was an error", but advising them to select a subscription is a good check

andrewki-msft: Okay, I have updated the user prompt as requested.

PR #187, opened 2019-05-13
Fix an issue with the locations tree - blank country/city causing iss…

acearun:
…ues.


acearun:
Query updated to name blank country/city entries to 'Unknown country' and 'Unknown city' respectively.

PR #192, opened 2019-05-15
Update at-scale Storage overview

acearun:
image


shijatsu:
Arun this is different from the template I have - please use my template to make changes

shijatsu:
1. it does not show all errors just 1 split by 
2. would need to show subscription name 
3. as we discussed with SA team availability will not have threshold 
4. same for latency - no threshold 
5. Details  title column is Egress timeline

PR #195, opened 2019-05-16
Removed threshold from transactions split by

shijatsu:


acearun:
You don't need this change anymore. Custom column settings are now persisted correctly.

PR #198, opened 2019-05-17
Some updates to the single storage templates

acearun:


Workbooks/Individual Storage/Table Overview/Table Overview.workbook
```diff
@@ -378,6 +344,7 @@
         "gridFormatting": []
       },
       "conditionalVisibility": null,
+      "showPin": true,
```
vutran01: do you also need to give other templates pinning capabilities?

PR #219, opened 2019-06-10
Users/paruizf/add custom endpoints docs

paruizfi:
Add docs from custom endpoint provider


Documentation/DataSources/DataSources.md
```diff
@@ -63,6 +64,16 @@ To make a query control use this data source, use the _Query Type_ drop down to
 
 ![A image of a workbook with Azure resource health data](../Images/ResourceHealthDataSource.png)
 
+## Custom endpoint
+Workbooks supports getting data from any external source. If your data lives outside Azure you can bring it to Workbooks by using this data source type.
+
+To make a query control use this data source, use the _Query Type_ drop down to choose _Custom Endpoint_. Provide the appropriate parameters such as Http method, url, headers, url parameters and/or body.
```
gardnerjr: there should probably be a `note` section here or something that indicates anything set in the headers/parameters/etc will be visible to any user of the workbook.  Until we have support for secrets, anything that requires api-keys or anything else will have any of those keys/etc visible in the workbook.

Documentation/DataSources/DataSources.md
```diff
@@ -63,6 +64,16 @@ To make a query control use this data source, use the _Query Type_ drop down to
...
+
+![A image of a workbook with Custom endpoint data](../Images/CustomEndpointsDataSource.png)
+
+To avoid automatically making calls to un-trusted hosts when using templates, you need to mark the used hosts as trusted. You can either do this by clicking on the _Add as trusted_ button, or by adding it as a trusted host in Workbook settings. This settings will be saved in browsers that support IndexDb with web workers, more info [here](https://caniuse.com/#feat=indexeddb).
```
gardnerjr: this says "you" need to mark the hosts.  The author reading here won't be able to do that.  this should probably say something like, when this workbook is run, "the *user* will need to..." because this is a user thing, not an authoring thing.

PR #226, opened 2019-06-17
Update at-scale storage template to have grouping and links

acearun:
image


Workbooks/Storage/Overview/Overview.workbook
```diff
@@ -16,9 +16,9 @@
             "name": "DefaultSubscription_Internal",
             "type": 1,
             "isRequired": true,
-            "query": "where type =~ 'microsoft.compute/virtualmachines'\r\n| take 1\r\n| project subscriptionId",
+            "query": "extend Rank = iff(type =~ 'microsoft.storage/storageaccounts', 1, 2)\r\n| order by Rank asc, subscriptionId asc\r\n| take 1\r\n| project subscriptionId",
             "crossComponentResources": [
-              "value::selected"
+              "value::all"
```
vutran01: are we sure with all instead of pick 3? all will take longer to query

acearun: I think this is between 'All subscriptions' and 'Default subscriptions'. I verified that the template defaults to Any 10 resources as expected. I don't think there will be a perf impact unless there are 1000s of subs.

PR #234, opened 2019-06-19
updated workbook

BLuEScioN:


gardnerjr:
@ericc1103 this is another PR that only shows the CLA check, i don't see it running the tests?

PR #267, opened 2019-06-29
done

BLuEScioN:


Workbooks/AKS/Node-Disk-Capacity-Cluster/settings.json
```diff
@@ -0,0 +1,6 @@
+{
+    "$schema": "https://github.com/Microsoft/Application-Insights-Workbooks/blob/master/schema/settings.json",
+    "name":"Node Disk Capacity",
+    "author": "Microsoft",
```
bragi92: no isPreview tag? We're directly releasing this?

PR #275, opened 2019-07-02
got rid of the f**** hyphens

BLuEScioN:


Workbooks/AKS/Node Disk Capacity AKS/Node Disk Capacity AKS.workbook
```diff
@@ -0,0 +1,142 @@
...
+47
+                  "durationMs": 43200000
+48
+                },
+@BLuEScioN
+Commit changes
+Commit summary 
+Update Node-Disk-Capacity-Cluster.workbook
+Optional extended description
+Add an optional extended description…
```
bragi92: should this be here?

PR #279, opened 2019-07-02
aks-engine workbook

bragi92:


Workbooks/AKS/Node-Disk-Capacity-AKS-Engine/settings.json
```diff
@@ -1,6 +1,6 @@
 {
     "$schema": "https://github.com/Microsoft/Application-Insights-Workbooks/blob/master/schema/settings.json",
-    "name":"Disk Capacity (Workspace)",
+    "name":"Disk Capacity (AKS-Engine)",
```
BLuEScioN: You don't have to differentiate. A different workbook blade with the corresponding workbooks should be loaded

bragi92: updated

PR #309, opened 2019-07-13
fix link to open right travis link, try to build master AND pr builds

gardnerjr:
I added the travis link to readme, and it shows the right icon for build status but clicking it went to Travis's docs instead of our builds


gardnerjr:
because of the travis yml change if this doesn't work i'm going to end up undoing that change in the next PR :D

PR #313, opened 2019-07-15
Update at-scale templates for the metric ids in renderers change

acearun:


ericajlin:
You might want to go into column label settings and add comments or at least click apply. Right now I see labelSettings as empty which happens when you don't edit the label settings and you choose to use the defaults, which is fine, but we will need to add it eventually for localization.

acearun:
> You might want to go into column label settings and add comments or at least click apply. Right now I see labelSettings as empty which happens when you don't edit the label settings and you choose to use the defaults, which is fine, but we will need to add it eventually for localization.

I'll change this with another PR.

PR #318, opened 2019-07-17
Add more at-scale templates for VMs

acearun:
1. Update at-scale metrics to use links that open up context blades.
2. Adds at-scale CPU heatmaps to both Azure Monitor and Resource Groups


acearun:
image

PR #325, opened 2019-07-23
Added template processing scripts

paruizfi:


scripts/ProcessTemplates.ps1
```diff
@@ -0,0 +1,143 @@
+$mainPath = Split-Path (split-path -parent $MyInvocation.MyCommand.Path) -Parent
```
gardnerjr: can you add some comments here about what this script does?

it should probably have input checks/etc that echo usage info and stuff if there's no args or whatever (although it looks like it requires no args and just uses the cwd as the arg?

scripts/ProcessTemplates.ps1
```diff
@@ -0,0 +1,149 @@
...
+#                 |- en
+#                 |   |- settings.json
+#                 |   |- readme.md
+#                 |   |- Apdex.workbook
+#                 |- ko
+#                     |- settings.json
+#                     |- readme.md
+#                     |- Apdex.workbook
+# 
+# If the transformation is successful the json gets published as artifact
```
gardnerjr: "if the transformation is successful, the json gets published as an artifact"

published where as what?

gardnerjr: does it just copy the file to a directory?  it doesn't go anywhere off your machine?

gardnerjr:
however you hook this up, make sure it only does *any* work when pushes to `master` and only `master` occur. one of the problems right now is that alm service gets told to update every time any push occurs, which means it updates for no reason when people create new branches for PR's and push to those branches

scripts/ProcessTemplates.ps1
paruizfi: the build has an artifacts folder that's where it gets published. Once the PR to master is completed a the release is triggered, the release definition uploads the artifact(JSON) to a storage account

gardnerjr: can you document this part somewhere:

> Once the PR to master is completed a the release is triggered, the release definition uploads the artifact(JSON) to a storage account

there's going to be a connection between this git repo and an VSTS pipeline or whatever, and how that is configured, etc.  here is just a script, but nothing in the *repo* uses the pipeline, some external pipeline does.

gardnerjr: can you document this part somewhere:

> Once the PR to master is completed a the release is triggered, the release definition uploads the artifact(JSON) to a storage account

there's going to be a connection between this git repo and an VSTS pipeline or whatever, and how that is configured, etc.  here is just a script, but nothing in the *repo* uses the pipeline, some external pipeline does.

PR #326, opened 2019-07-23
Update step names for telemetry and add download to excel buttons

acearun:


vutran01:
Could’ve use margins on some of these

PR #330, opened 2019-07-26
Adding Device Compliance Trends to Intune Reports

RoyceMou:
image

### DIFF
--- a/Workbooks/Intune Reports/Device Compliance Trends/DeviceComplianceTrends.workbook
+++ b/Workbooks/Intune Reports/Device Compliance Trends/DeviceComplianceTrends.workbook
@@ -1,0 +1,82 @@
+{
+  "version": "Notebook/1.0",
+  "items": [
+    {
+      "type": 1,
+      "content": {
+        "json": "Device Compliance\r\n=="
+      },
+      "name": "text - 2"
+    },
+    {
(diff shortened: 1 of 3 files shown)


Workbooks/Intune Reports/Device Compliance Trends/DeviceComplianceTrends.workbook
```diff
@@ -0,0 +1,82 @@
+{
+  "version": "Notebook/1.0",
+  "items": [
+    {
+      "type": 1,
+      "content": {
+        "json": "Device Compliance\r\n=="
```
gardnerjr: instead of doing this as a top level text item, you _might_ want to set this as the title of the chart itself (see advanced settings of that chart, there's a field for title there)  but for the most part this is a style thing so it's really up to you.

the chart title field (falling back to the step name field) are used for accessibility too, so we suggest setting at least step name to something useful, otherwise they end up with autogenerated names like "text - 2"

RoyceMou: Thanks for the info! I'll have to follow up with a PR to figure out the rest of the OS filter values once we figure that out so I'll make the step name changes together then.

PR #334, opened 2019-07-30
Storage Insights: Add resource limit

ericajlin:
Add max resource limit of 200

### DIFF
--- a/Workbooks/Storage/Overview/Overview.workbook
+++ b/Workbooks/Storage/Overview/Overview.workbook
@@ -502,6 +502,7 @@
           }
         ],
         "filters": [],
+        "resourceLimit": 200,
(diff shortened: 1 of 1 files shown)


Workbooks/Storage/Overview/Overview.workbook
```diff
@@ -502,6 +502,7 @@
           }
         ],
         "filters": [],
+        "resourceLimit": 200,
```
gardnerjr: isn't 200 the default?  or is there no more default?

ericajlin: I put a default limit of 200 in the code and I thought that it worked but apparently not. I'll have to fix it

PR #338, opened 2019-07-31
Storage Insights: Add missing step names

ericajlin:
Give more meaningful step names to storage insights templates so we can track telemetry easier

### DIFF
--- a/Workbooks/Individual Storage/Overview/Overview.workbook
+++ b/Workbooks/Individual Storage/Overview/Overview.workbook
@@ -335,7 +335,7 @@
           "showBorder": false
         }
       },
-      "name": "query - 18"
+      "name": "Overview tiles"
     },
     {
       "type": 1,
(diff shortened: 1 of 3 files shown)


Workbooks/Individual Storage/Overview/Overview.workbook
```diff
@@ -335,7 +335,7 @@
           "showBorder": false
         }
       },
-      "name": "query - 18"
+      "name": "Overview tiles"
```
gardnerjr: isn't "overview" in the title of the workbook?  might be overkill to have it in every step here, then you'll have really long names in columns an the beginning will always be "overview " and useless in grids and stuff?

personally, i'd trim them down to the minimum useful thing to distinguish steps

ericajlin: these steps don't have references in the rest of the workbooks though. I'm not sure what particular overhead that would cost? I put overview here to add distinguish in telemetry between the steps we have in storage insights (We have like 5 templates all showing stuff like tiles, latency, transactions).

gardnerjr: the template names are already in telemetry in another field though.  so you'll see
overview in all the names in a column like this:
image

PR #343, opened 2019-08-02
Storage Insights: Resource Overview should take time param

ericajlin:
Summary step was defaulting to 1 hour instead of using the time parameter


ericajlin:
image

PR #350, opened 2019-08-08
Initial version of SAP Monitor template

tniek:
- Add file/folder structure
- Add basic template with utilization and forecast metrics
- Add `{LogAnalyticsWorkspaceArmId}` placeholder

### DIFF
--- a/Workbooks/SapMonitor/SapHanaInfrastructure/SapHanaInfrastructure.workbook
+++ b/Workbooks/SapMonitor/SapHanaInfrastructure/SapHanaInfrastructure.workbook
@@ -175,7 +175,6 @@
         "queryType": 0,
         "resourceType": "microsoft.operationalinsights/workspaces",
         "crossComponentResources": [
-          "{LogAnalyticsWorkspaceArmId}"
         ],
         "visualization": "areachart"
       },
@@ -198,7 +197,6 @@
         "queryType": 0,
         "resourceType": "microsoft.operationalinsights/workspaces",
         "crossComponentResources": [
-          "{LogAnalyticsWorkspaceArmId}"
         ],
         "visualization": "areachart"
       },
@@ -220,7 +218,6 @@
         "queryType": 0,
         "resourceType": "microsoft.operationalinsights/workspaces",
         "crossComponentResources": [
-          "{LogAnalyticsWorkspaceArmId}"
         ],
         "visualization": "categoricalbar"
       },
@@ -243,7 +240,6 @@
         "queryType": 0,
         "resourceType": "microsoft.operationalinsights/workspaces",
         "crossComponentResources": [
-          "{LogAnalyticsWorkspaceArmId}"
         ],
         "visualization": "areachart"
       },
@@ -265,7 +261,6 @@
         "queryType": 0,
         "resourceType": "microsoft.operationalinsights/workspaces",
         "crossComponentResources": [
-          "{LogAnalyticsWorkspaceArmId}"
         ],
         "visualization": "linechart",
         "graphSettings": {
@@ -312,7 +307,6 @@
         "queryType": 0,
         "resourceType": "microsoft.operationalinsights/workspaces",
         "crossComponentResources": [
-          "{LogAnalyticsWorkspaceArmId}"
         ],
         "visualization": "linechart",
         "graphSettings": {
@@ -337,7 +331,6 @@
         "queryType": 0,
         "resourceType": "microsoft.operationalinsights/workspaces",
         "crossComponentResources": [
-          "{LogAnalyticsWorkspaceArmId}"
         ],
         "visualization": "linechart"
       },
@@ -359,7 +352,6 @@
         "queryType": 0,
         "resourceType": "microsoft.operationalinsights/workspaces",
         "crossComponentResources": [
-          "{LogAnalyticsWorkspaceArmId}"
         ],
         "visualization": "linechart"
       },
--- a/Workbooks/SapMonitor/SapHanaInfrastructure/SapHanaInfrastructure.workbook
+++ b/Workbooks/SapMonitor/SapHanaInfrastructure/SapHanaInfrastructure.workbook
@@ -10,7 +10,6 @@
         "queryType": 0,
         "resourceType": "microsoft.operationalinsights/workspaces",
         "crossComponentResources": [
-          "{LogAnalyticsWorkspaceArmId}"
         ],
         "visualization": "table",
         "gridSettings": {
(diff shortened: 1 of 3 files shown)


Workbooks/SapMonitor/SapHanaInfrastructure/SapHanaInfrastructure.workbook
```diff
@@ -0,0 +1,376 @@
...
+    {
+      "type": 3,
+      "content": {
+        "version": "KqlItem/1.0",
+        "query": "SapHana_LoadHistory_CL\r\n|summarize arg_max(UTC_TIMESTAMP_t, *) by HOST_s\r\n|project Host=HOST_s , [\"CPU Utilization (%)\"]=CPU_d ,  [\"Memory Utilization (%)\"]=round((MEMORY_USED_d/MEMORY_SIZE_d)*100,2),[\"Disk Utilization (GB)\"]=DISK_USED_d  , [\"Network In\"]=NETWORK_IN_d , [\"Network Out\"]=NETWORK_OUT_d ",
+        "size": 4,
+        "queryType": 0,
+        "resourceType": "microsoft.operationalinsights/workspaces",
+        "crossComponentResources": [
+          "{LogAnalyticsWorkspaceArmId}"
```
gardnerjr: you have this as a parameter here, but there are no parameters steps.

if this workbook is opened by your blade, and your blade passes in the workspace as `ResourceIds: [ "that workspace id" ]` in the blade parameters, all the query steps will "inherit" that workspace as the default resource.

in that case, you could just leave crossComponentResources empty here in all the steps.

otherwise, one way would be to add a parameters step at the top, and declare a param of type "resource picker", with that name, and have it get its value from the workbook's resources.  then this parameter would be valid here with that name.

Workbooks/SapMonitor/SapHanaInfrastructure/SapHanaInfrastructure.workbook
```diff
@@ -0,0 +1,376 @@
...
+      },
+      "name": "text - 1"
+    },
+    {
+      "type": 3,
+      "content": {
+        "version": "KqlItem/1.0",
+        "query": "SapHana_LoadHistory_CL\r\n| summarize avg(CPU_d) by HOST_s , bin(UTC_TIMESTAMP_t, 15m)",
+        "size": 0,
+        "timeContext": {
```
gardnerjr: it looks like some of the steps have an explicit time range set, and others are using "set in query", but then don't actually have a time range filter in the query.  without an explicit time range set in the query, it will default to last 24h.

you might want to add a time range parameter at the top and have all the queries use that time range parameter?

Workbooks/SapMonitor/SapHanaInfrastructure/SapHanaInfrastructure.workbook
```diff
@@ -0,0 +1,376 @@
...
+        "resourceType": "microsoft.operationalinsights/workspaces",
+        "crossComponentResources": [
+          "{LogAnalyticsWorkspaceArmId}"
+        ],
+        "visualization": "linechart"
+      },
+      "name": "query - 10 - Copy"
+    }
+  ],
+  "fallbackResourceIds": [
```
gardnerjr: the resources down here are the thing the PR is rejecting.  in a template, you can remove this whole section

Workbooks/SapMonitor/SapHanaInfrastructure/SapHanaInfrastructure.workbook
```diff
@@ -0,0 +1,376 @@
...
+        "resourceType": "microsoft.operationalinsights/workspaces",
+        "visualization": "linechart",
+        "graphSettings": {
+          "type": 0
+        }
+      },
+      "name": "query - 7 - Copy"
+    },
+    {
+      "type": 1,
```
gardnerjr: you're using a lot of text steps as titles.  the query steps can have a title instead, see chart title in advanced settings in the query step.

Workbooks/SapMonitor/SapHanaInfrastructure/SapHanaInfrastructure.workbook
```diff
@@ -0,0 +1,376 @@
...
+        "resourceType": "microsoft.operationalinsights/workspaces",
+        "crossComponentResources": [
+          "{LogAnalyticsWorkspaceArmId}"
+        ],
+        "visualization": "linechart",
+        "graphSettings": {
+          "type": 0
+        }
+      },
+      "name": "query - 7 - Copy - Copy"
```
gardnerjr: while not *necessary*, we also suggest setting step titles in the individual steps in the template as well. the step names are useful for people who might be editing them later, and the step names are used in telemetry to identify which queries in your template are passing/failing/taking time.  we're working on exposing query performance telemetry as well, see the wiki in teams for onboarding instructions there.

Workbooks/SapMonitor/SapHanaInfrastructure/SapHanaInfrastructure.workbook
```diff
@@ -0,0 +1,376 @@
+{
+  "version": "Notebook/1.0",
+  "items": [
+    {
+      "type": 3,
+      "content": {
+        "version": "KqlItem/1.0",
+        "query": "SapHana_LoadHistory_CL\r\n|summarize arg_max(UTC_TIMESTAMP_t, *) by HOST_s\r\n|project Host=HOST_s , [\"CPU Utilization (%)\"]=CPU_d ,  [\"Memory Utilization (%)\"]=round((MEMORY_USED_d/MEMORY_SIZE_d)*100,2),[\"Disk Utilization (GB)\"]=DISK_USED_d  , [\"Network In\"]=NETWORK_IN_d , [\"Network Out\"]=NETWORK_OUT_d ",
```
gardnerjr: also, is `UTC_TIMESTAMP_t` the standard time column in your tables?  the code that figures out if a time range is specified in a query is looking for specific known column names (like `timestamp` in app insights and `TimeGenerated` in most log analytics tables.  do i need to add this to our list?

Workbooks/SapMonitor/SapHanaInfrastructure/SapHanaInfrastructure.workbook
tniek: Great, I'll just leave the `crossComponentResources` section empty then.

Workbooks/SapMonitor/SapHanaInfrastructure/SapHanaInfrastructure.workbook
tniek: Sure; the current template is really just a "placeholder" to test the integration.
Will follow your suggested design patterns and add a time range picker at the top.

Workbooks/SapMonitor/SapHanaInfrastructure/SapHanaInfrastructure.workbook
tniek: Will build that into the "real" template.

Workbooks/SapMonitor/SapHanaInfrastructure/SapHanaInfrastructure.workbook
tniek: As I'm batch-ingesting the collected metrics using Log Analytics's [Data Collector API](https://docs.microsoft.com/en-us/azure/azure-monitor/platform/data-collector-api), I went for a separate custom field `UTC_TIMESTAMP_t` for each record. (I understand there is a way to overwrite the TimeGenerated field by setting the `time-generated-field` HTTP header, but that would limit me to one row per HTTP ingestion call.)

So yes, please add UTC_TIMESTAMP_t to your list.

PR #356, opened 2019-08-14
changed from All sub to default sub

shijatsu:
attempt to handle too many sub scenario


Workbooks/CosmosDb/Overview/Overview.workbook
```diff
@@ -7,7 +7,7 @@
         "version": "KqlParameterItem/1.0",
         "query": "",
         "crossComponentResources": [
-          "{Subscription}"
+          "value::selected"
```
gardnerjr: this part is interesting, this section should be "empty" for parameters steps.  yet another little thing to look into

shijatsu: yup I noticed that

PR #369, opened 2019-08-22
Storage Insights: Use column labels

ericajlin:
Column name option has been removed from metric pills and moved to column label options in settings blade. Remove column name from each metric setting and move it to column label settings.

Removes the confusion where someone tries to change Availability metric  to say Egress metric but the column name still says Availability.

### DIFF
--- a/Workbooks/Storage/Capacity/Capacity.workbook
+++ b/Workbooks/Storage/Capacity/Capacity.workbook
@@ -246,46 +246,31 @@
             "namespace": "microsoft.storage/storageaccounts",
             "metric": "microsoft.storage/storageaccounts-Capacity-UsedCapacity",
             "aggregation": 4,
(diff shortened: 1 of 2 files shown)


Workbooks/Storage/Capacity/Capacity.workbook
```diff
@@ -246,46 +246,31 @@
             "namespace": "microsoft.storage/storageaccounts",
             "metric": "microsoft.storage/storageaccounts-Capacity-UsedCapacity",
             "aggregation": 4,
-            "splitBy": null,
-            "splitBySortOrder": -1,
```
ericajlin: Someone probably tried to add a split by here, changed it back to no split by, but we don't check if the split by is null before saving the setting. I should refactor this in code to avoid unnecessary saved settings

PR #375, opened 2019-08-30
SAP Monitor infrastructure template v2

tniek:
- Cleaned up JSON (assigned meaningful step names)
- Added tabs for utilization, anomalies and host config
- Improved labels in KQL queries

### DIFF
--- a/Workbooks/SapMonitor/SapHanaInfrastructure/SapHanaInfrastructure.workbook
+++ b/Workbooks/SapMonitor/SapHanaInfrastructure/SapHanaInfrastructure.workbook
@@ -2,6 +2,34 @@
   "version": "Notebook/1.0",
   "items": [
     {
+      "type": 9,
+      "content": {
+        "version": "KqlParameterItem/1.0",
+        "query": "",
+        "crossComponentResources": [],
+        "parameters": [
+          {
+            "id": "386c5637-375e-437c-acdb-542be563ed5a",
+            "version": "KqlParameterItem/1.0",
+            "name": "testDataExist",
+            "type": 1,
+            "query": "SapHana_LoadHistory_CL\r\n| where TimeGenerated  {TimeRange}\r\n|project HOST_s\r\n| take 1",
+            "isHiddenWhenLocked": true,
+            "timeContext": {
+              "durationMs": 0
+            },
+            "timeContextFromParameter": "TimeRange",
+            "queryType": 0,
+            "resourceType": "microsoft.operationalinsights/workspaces"
+          }
+        ],
+        "style": "pills",
+        "queryType": 0,
+        "resourceType": "microsoft.operationalinsights/workspaces"
+      },
+      "name": "testDataExist_Parameters"
+    },
+    {
       "type": 11,
       "content": {
         "version": "LinkItem/1.0",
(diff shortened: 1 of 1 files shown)


Workbooks/SapMonitor/SapHanaInfrastructure/SapHanaInfrastructure.workbook
```diff
@@ -1,17 +1,649 @@
...
+          {
+            "cellValue": "selectedTab",
+            "linkTarget": "parameter",
+            "linkLabel": "Host Config",
+            "subTarget": "3",
+            "style": "secondary"
+          }
+        ]
+      },
+      "conditionalVisibility": {
```
gardnerjr: the first step of this workbook appears to have a conditionalVisibility setting that makes it always visible?  might want to just delete that part?

Workbooks/SapMonitor/SapHanaInfrastructure/SapHanaInfrastructure.workbook
```diff
@@ -1,17 +1,649 @@
...
+                  "durationMs": 5184000000
+                }
+              ]
+            },
+            "label": "Time range"
+          },
+          {
+            "id": "fae26122-3768-43dd-ac98-9b9768924799",
+            "version": "KqlParameterItem/1.0",
+            "name": "Granularity",
```
gardnerjr: if you don't want to force the user to do this to pick a grain, you can use `{TimeRange:grain}` in the query to get a "reasonable" default grain based on the selected time range

Workbooks/SapMonitor/SapHanaInfrastructure/SapHanaInfrastructure.workbook
```diff
@@ -1,17 +1,649 @@
...
+            "jsonData": "[\r\n    { \"value\":\"1m\", \"label\":\"1 Minute\", \"selected\":false},\r\n{ \"value\":\"5m\", \"label\":\"5 Minutes\", \"selected\":false},\r\n{ \"value\":\"15m\", \"label\":\"15 Minutes\", \"selected\":true},\r\n{ \"value\":\"30m\", \"label\":\"30 Minutes\", \"selected\":false},\r\n{ \"value\":\"1h\", \"label\":\"1 Hour\", \"selected\":false}\r\n]\r\n",
+            "timeContext": {
+              "durationMs": 0
+            },
+            "timeContextFromParameter": "TimeRange"
+          },
+          {
+            "id": "8e3577a4-1f9f-4935-a2b2-c42a6911ae8a",
+            "version": "KqlParameterItem/1.0",
+            "name": "Host",
```
gardnerjr: this param doesn't seem to be an explicit time range in the query or using a time range parameter, so it might be implicitly getting 24h as the time range.  it should probably be using the time range parameter?

gardnerjr:
without having real data to try it on, can you paste a screenshot or something just for the general gist?

tniek:
> without having real data to try it on, can you paste a screenshot or something just for the general gist?

image

Workbooks/SapMonitor/SapHanaInfrastructure/SapHanaInfrastructure.workbook
tniek: Actually, the condition testDataExist != (empty) checks if the ingested data has already "arrived" to Log Analytics. We use it pretty much in every step and display an error message if the data is not there yet.

Workbooks/SapMonitor/SapHanaInfrastructure/SapHanaInfrastructure.workbook
tniek: These "Host Config" metrics are supposed to show the current (latest) status only, so the time range shouldn't apply here.

Workbooks/SapMonitor/SapHanaInfrastructure/SapHanaInfrastructure.workbook
tniek: Thanks, noted. I think for the time being we would expect users to specify the granularity (same behavior as SAP HANA Cockpit).

Workbooks/SapMonitor/SapHanaInfrastructure/SapHanaInfrastructure.workbook
gardnerjr: yes, but this is the tabs very first step of the workbook, which exists before that parameter exists?

tniek: Good catch - fixed

PR #389, opened 2019-09-20
Remove VM Insights workbooks from workspace gallery

andrewki-msft:
I pushed these same changes to https://github.com/ericc1103/Application-Insights-Templates-Dev but for some reason the Log Analytics workbook gallery is still showing VM Insights templates. Will take a look again tomorrow.

image


andrewki-msft:
I'm going ahead and pushing this in since it's a relatively safe change and there appears to be some blockage on INT.

PR #390, opened 2019-09-20
cosmosDB insights capacity changes

heyaWorld:
New CosmosDB Capacity View:

cosmos new capacity view

palette changes, dropdown option changes, enable export to excel.


heyaWorld:
> the indented tabs seem odd to me

yep, removed it

PR #411, opened 2019-10-07
Remove unnecessary KQL comments

andrewki-msft:
KQL comments were originally introduced for better loading user experience. Recent updates in workbooks makes adding such KQL comments unnecessary. Also KQL queries were causing problems when the query strings were getting too long.

AtScale
image

SingleVM
image

### DIFF
--- a/Workbooks/Virtual Machines - Network Dependencies/Connections Overview - Single VM/Connections Overview.workbook
+++ b/Workbooks/Virtual Machines - Network Dependencies/Connections Overview - Single VM/Connections Overview.workbook
@@ -6,7 +6,6 @@
       "content": {
         "json": "<h1>Connections Overview</h1>"
       },
-      "conditionalVisibility": null,
       "name": "text - 0"
     },
     {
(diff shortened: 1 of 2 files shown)


Workbooks/Virtual Machines - Network Dependencies/Connections Overview - Single VM/Connections Overview.workbook
```diff
@@ -491,76 +453,69 @@
             "name": "QueryPadProcessNameTable",
             "type": 1,
             "isRequired": true,
-            "query": "// {Computer:label}\r\nprint(\"let processNamePadding = datatable (ProcessName: string) [];\");",
+            "query": "print(\"let processNamePadding = datatable (ProcessName: string) [];\");",
```
gardnerjr: why are some of these queries?  am I missing something?  this looks like just a static string that could be part of a different query?  it doesn't have any params, so there's no replacements going on.  am I missing something?

Workbooks/Virtual Machines - Network Dependencies/Connections Overview - Single VM/Connections Overview.workbook
```diff
@@ -727,7 +682,77 @@
             "treeType": 0,
             "expanderColumn": "Name",
             "expandTopLevel": true
-          }
+          },
+          "labelSettings": [
+            {
+              "columnId": "ComputerName",
+              "label": "ComputerName"
```
gardnerjr: did you explicitly add these?  if so, the labels should be nice names, with spaces and things, like "Computer Name".  right now, they're all the same as the keys and this is probably a bug @ericajlin ?

andrewki-msft: Hi John, I did not explicitly add these. The template itself is 6 month old so lots of new metadata is added to the template by default. What I typically do is load the template, make changes, and manually remove any hard-coded entries and update the PR. I will go back and update some of these labels.

Workbooks/Virtual Machines - Network Dependencies/Connections Overview - Single VM/Connections Overview.workbook
andrewki-msft: Thanks for catching this. Removed.

PR #414, opened 2019-10-07
Adding workbook templates for Azure AD Domain Services

MikeStephens-MS:
These are two Azure Workbook templates (Security Overview Report and Account Activity Report) that are used with Azure AD Domain Services (not to be confused with Azure Active Directory.  

The Security Overview Report helps customers with reviewing the security audit logs that come from Azure AD Domain Services. Using this templates, customers can see historical views of sign-ins and the distribution of sign-in results (bad password, account disabled, account locked out, etc).  Additionally, the report highlights locked out users (are biggest support generator)--enabling customers to track repeated bad password sign ins and the subsequently lockout. The report also helps customers identify bad usernames that attackers use when attempting password spray attacks either through RDP or over secure LDAP.  
Capture4
Capture8

Our Account Activity Report provides a timeline view of events with anomaly detection to help customers focus on the potential threats while also seeing the breadth of events as they occur; making it easy to form baselines.
Capture2


MikeStephens-MS:
@gardnerjr -- can you or someone review this and accept the PR into master?  Many thanks,

PR #416, opened 2019-10-07
add missing schema to example

gardnerjr:
add missing schema to settings.json example
add generated schema for settings.json


schema/settings.json
```diff
@@ -0,0 +1,137 @@
+{
```
acearun: Nice!

gardnerjr: (yeah, this one is easy to generate, workbooks not so much)

PR #419, opened 2019-10-08
AtScale Perf Preview Workbook

andrewki-msft:
Uses InsightsMetrics table, for both Context and Context-free modes

Context-free mode
image

AtScale VMI
image


gardnerjr:
do you want to split it into tabs?  one tab for top 100, one for top 10?  then all the top10 queries don't run until you pick that tab?

gardnerjr:
nit: you no longer need the splitter above the tabs, and you don't need the text "top X machines" inside each tab, as the tabs do that work for you.  would get some space back to show the actual charts

PR #430, opened 2019-10-12
Updating agent health template.

avidhanju:
Updating agent health template and adding at scale template for azure monitor.

Screenshots:
1. With data
image

2. Without data
image


Workbooks/Azure Monitor - Workspaces/Agent Health/settings.json
```diff
@@ -0,0 +1,6 @@
+{
+    "$schema": "https://github.com/Microsoft/Application-Insights-Workbooks/blob/master/schema/settings.json",
+    "name":"Workspace Usage",
```
dougbrad: Name is wrong ?

Workbooks/Azure Monitor - Workspaces/Agent Health/Agent Health.workbook
```diff
@@ -0,0 +1,615 @@
...
+              "columnId": "OS",
+              "label": "OS"
+            },
+            {
+              "columnId": "Azure Resource",
+              "label": "Azure Resource"
+            },
+            {
+              "columnId": "Time",
+              "label": "Time"
```
dougbrad: It would be more intuitive if this column named something like 'Last Heartbeat' instead of 'Time'

Workbooks/Azure Monitor - Workspaces/Agent Health/Agent Health.workbook
```diff
@@ -0,0 +1,615 @@
...
+            "queryType": 1,
+            "resourceType": "microsoft.resourcegraph/resources"
+          },
+          {
+            "id": "20fb47ba-f521-47ea-9fb7-66339ad9c56a",
+            "version": "KqlParameterItem/1.0",
+            "name": "Subscription",
+            "type": 6,
+            "isRequired": true,
+            "query": "summarize by subscriptionId\r\n| project value = strcat(\"/subscriptions/\", subscriptionId), label = subscriptionId, selected = iff(subscriptionId =~ '{DefaultSubscription_Internal}', true, false)\r\n",
```
dougbrad: Shall we only list Subscriptions that have Workspaces? Actually wonder if a Subscription picker really needed... Just showing all workspaces may be ok, most customers have a reasonable # of these, at least in default subscriptions.

Workbooks/Azure Monitor - Workspaces/Agent Health/Agent Health.workbook
```diff
@@ -0,0 +1,615 @@
...
...romNow < 2h, strcat(toint(TimeFromNow / 1m), ' minutes'), TimeFromNow < 2d, strcat(toint(TimeFromNow / 1h), ' hours'), strcat(toint(TimeFromNow / 1d), ' days')), ' ago')\r\n| join (\r\nHeartbeat\r\n| where TimeGenerated {TimeRange:query}\r\n| extend Packed = pack_all()\r\n) on Computer\r\n| where TimeGenerated == LastHeartbeat\r\n| join (\r\nHeartbeat\r\n| where TimeGenerated {TimeRange:query}\r\n| make-series InternalTrend=iff(count() > 0, 1, 0) default = 0 on TimeGenerated from {TimeRange:start} to {TimeRange:end} step {UnhealthyCriteria} by Computer\r\n| extend Trend=array_slice(InternalTrend, array_length(InternalTrend) - 30, array_length(InternalTrend)-1)\r\n| extend (s_min, s_minId, s_max, s_maxId, s_avg, s_var, s_stdev) = series_stats(Trend)\r\n| project Computer, Trend, s_avg\r\n) on Computer\r\n| order by State, s_avg asc, TimeAgo\r\n| project [\"_ComputerName_\"] = Computer, [\"Computer\"]=strcat('🖥️ ', Computer), State, [\"Environment\"] = iff(ComputerEnvironment == \"Azure\", ComputerEnvironment, Category), [\"OS\"]=iff(isempty(OSName), OSType, OSName), [\"Azure Resource\"]=ResourceId, [\"Time\"]=strcat('🕒 ', TimeAgo), [\"Heartbeat Trend\"]=Trend, [\"Details\"]=Packed",
```
dougbrad: I wonder on these queries in large environments w/ thousands of agents, for example we are doing:

    Heartbeat
    | where TimeGenerated > ago(30d)
    | extend Packed = pack_all()

Can we instead use arg_min so we don't have to fetch and pack all events in the 30 days?

Workbooks/Azure Monitor - Workspaces/Agent Health/Agent Health.workbook
avidhanju: Maybe. But for now I am simply following what we are doing in all other templates at Azure Monitor level.

Workbooks/Azure Monitor - Workspaces/Agent Health/Agent Health.workbook
avidhanju: We are doing that. We are taking `LastHeartbeat = max(TimeGenerated)` and joining it with the entire Heartbeat table on Computer name as the key and the filtering down to that last time range by using `| where TimeGenerated == LastHeartbeat`. So in essence, we are only doing pack_all on only the last heartbeat of each computer.

PR #432, opened 2019-10-14
Add tabs to at-scale storage template

acearun:


acearun:
image

gardnerjr:
now that we're making other things bigger/tabs, does the numbers line need to be so big?

# 10/10
### storage accounts


vs just

### 10/10 storage accounts

(the formatting won't be exactly the same here but the idea still stands)

that number and text seems to take a up a LOT of valuable real estate in the report

acearun:
@gardnerjr, interesting idea. Will consider as part of a separate change.

PR #444, opened 2019-10-18
generate template json for localized lang

vutran01:
generate the following json files from the localized repos

community-templates-V2.json
community-templates-V2.cs-cz.json
community-templates-V2.en-us.json
community-templates-V2.es-es.json
community-templates-V2.fr-fr.json
community-templates-V2.hu-hu.json
community-templates-V2.it-it.json
community-templates-V2.ja-jp.json
community-templates-V2.ko-kr.json
community-templates-V2.nl-nl.json
community-templates-V2.pl-pl.json
community-templates-V2.pt-br.json
community-templates-V2.ru-ru.json
community-templates-V2.sv-se.json
community-templates-V2.tr-tr.json
community-templates-V2.zh-cn.json
community-templates-V2.zh-tw.json


scripts/ProcessTemplates.ps1
```diff
@@ -1,24 +1,50 @@
...
 $payload = @{ }
 $categoryMetadataFileName = 'categoryResources.json'
 
 # This is the name of where the blob that ALM looks for
-$azureBlobFilename = "community-templates-V2.json";
-
+$azureBlobFileNameBase = "community-templates-V2";
+
+$docRepos = @(
+    @{ repo="Application-Insights-Workbooks.cs-cz"; lang="cs-cz"},
```
gardnerjr: do you need the language of the repo separate from the name of the repo?  it seems like passing this info 2x is a super way to be wrong?  instead would it make more sense to remove it from the name of the repo here, and just append it in the code when `lang !== "en-us"` ?

scripts/ProcessTemplates.ps1
```diff
@@ -160,94 +201,234 @@ function Copy-Object {
...
+#----------------------------------------------------------------------------
+# CloneLocalizedRepos
+#----------------------------------------------------------------------------
+Function CloneLocalizedRepos {
+    $localSrc = Convert-Path "..\.."
+
+    foreach ($repo in $docRepos) {
+        $currentRepo = $repo.repo
+        if (Test-Path $localSrc$currentRepo) {
+            Write-Host "Trying to clone repo $currentRepo but it already exist on disk, skipping"
```
gardnerjr: if the repo already exists on the disk wouldn't you want to pull to get the up to date version?
or is the code guaranteed not to be on the machine when the build starts and this can get called multiple times?

scripts/ProcessTemplates.ps1
vutran01: simplify the languages and use just the lang

scripts/ProcessTemplates.ps1
vutran01: doing git pull if exist

PR #449, opened 2019-10-22
CosmosDB Insights step names + expand button

ericajlin:
Add cosmos dbi step names, and add expand/collapse grid toolbar items


ericajlin:
Not sure why there's weird space formatting for "showExpandCollapseGrid". Even when I delete spacing it still shows up in the diff. Still works though

PR #464, opened 2019-10-24
Changed the queries to use InsightsMetrics for Performance Counter Workbook

rakshith210:


andrewki-msft:
@rakshith210 would you add some screenshots

rakshith210:
perfCountersWorkbook

rakshith210:
> @rakshith210 would you add some screenshots

Done

andrewki-msft:
Few things, but we can take care of this before the template goes public

- Time range should be set to 3 days, for some reason I'm not seeing any data with 48 hours
- Should be more pre-set counter variety, right now I see a lot of counters being repeated in the 2nd row

PR #465, opened 2019-10-24
Update key vault templates

acearun:
image

image

image


Workbooks/KeyVault/Azure Monitor Onboarding/Azure Monitor Onboarding.workbook
```diff
@@ -0,0 +1,281 @@
...
+                  "text": ""
+                }
+              ]
+            }
+          },
+          "secondaryContent": {
+            "columnMatch": "Footer",
+            "formatter": 1,
+            "formatOptions": {
+              "linkTarget": "Url",
```
dougbrad: Just to make sure - this makes a 'Change' link show on the tile which goes nowhere for me.
It will work once deployed?

Workbooks/KeyVault/Azure Monitor Onboarding/Azure Monitor Onboarding.workbook
```diff
@@ -0,0 +1,281 @@
...
+      },
+      "name": "query - 4",
+      "styleSettings": {
+        "margin": "10px 0 0 0"
+      }
+    },
+    {
+      "type": 3,
+      "content": {
+        "version": "KqlItem/1.0",
```
dougbrad: Ideally we could filter this table based on if the tile is clicked (Enabled or Disabled)

acearun: Makes sense - we can do it in the next version.

PR #469, opened 2019-10-25
Changed the query in VMSS Perf Analysis to use InsightsMetrics

rakshith210:
Changed the charts in Perf Analysis workbook for VMSS to use InsightsMetrics

image

### DIFF
--- a/Workbooks/Virtual Machine Scale Sets - Performance Analysis (Preview)/Virtual Machine Scale Sets - Performance Analysis (Preview).workbook
+++ b/Workbooks/Virtual Machine Scale Sets - Performance Analysis (Preview)/Virtual Machine Scale Sets - Performance Analysis (Preview).workbook
@@ -210,9 +210,39 @@
       "name": "text - 4"
     },
     {
+      "type": 11,
+      "content": {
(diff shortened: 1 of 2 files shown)


gardnerjr:
are those 4 charts always those 4 metrics?  if so, use chart settings to set y axis range 0-100 for the % one, set the y axis units on all of them to the appropriate things?  then the metrics at the bottom, labels on the side will show the right info?

gardnerjr:
also, I've previously suggested making this into tabs so that only the top 10 or top 100 sections are showing, which would prevent some of the queries from running all the time?

rakshith210:
> also, I've previously suggested making this into tabs so that only the top 10 or top 100 sections are showing, which would prevent some of the queries from running all the time?

I made the change to create tabs for top 10 and top 100 sections. Below are the screenshots of the same.

rakshith210:
workbookTopMachines10Tab

workbookTopMachinesTab

gardnerjr:
same comment on the at scale one andrew just did:  you can get rid of the gray divider lines and text inside the top of each tab and get some space back.  the tab control itself now is that divider and text

PR #470, opened 2019-10-26
First iteration Workbook Schema

ericajlin:
First version of workbook schema, includes fields needed for localization.

- Missing details for parameters, query steps, format settings

Other comments:
- To use this schema to do validation in advanced editor, it seems to me that the uri property no longer works (search JSON editor in internal github) and the whole schema needs to be passed into the viewmodel (best if it had it's own file and then reference it)
- I can't seem to get the schema to validate type other than the first level properties. Not sure what I'm missing?


schema/workbook.json
```diff
@@ -1,2 +1,754 @@
...
+            "$id": "#/properties/fallbackResourceIds",
+            "type": "array",
+            "description": "resourceIds that are opened with the workbook. Used if the workbook is loaded later from a place * with no explicit context passed in (like from a share link or a pin)"
+        },
+        "$schema": {
+            "$id": "#/properties/$schema",
+            "type": "string",
+            "title": "$schema",
+            "default": "https://github.com/Microsoft/Application-Insights-Workbooks/blob/master/schema/workbook.json",
+            "examples": [
```
gardnerjr: for schema, isn't just `default` field good enough`?  shouldn't need any example/pattern as isn't this technically the only legit value?

schema/workbook.json
```diff
@@ -1,2 +1,754 @@
...
+                "type": {
+                    "const": 1
+                },
+                "content": {
+                    "type": "object",
+                    "properties": {
+                        "json": {
+                            "type": "string",
+                            "examples": [
+                                "<h1>This is some markdown text</h1>"
```
gardnerjr: nit, this isn't markdown text, this is html. maybe want `# this is some markdown text` ?

schema/workbook.json
```diff
@@ -1,2 +1,754 @@
...
+                    "const": 11
+                },
+                "content": {
+                    "type": "object",
+                    "properties": {
+                        "version": {
+                            "type": "string",
+                            "description": "Version of link step",
+                            "examples": [
+                                "LinkItem/2.0"
```
gardnerjr: some of these should just be default `version` not examples?  there's technically only one allowed version and its version is `LinkItem/1.0`

schema/workbook.json
```diff
@@ -1,2 +1,754 @@
...
+                            ]
+                        },
+                        "style": {
+                            "type": "string",
+                            "description": "style of link",
+                            "examples": [
+                                "bullets",
+                                "paragraph",
+                                "navigation"
```
gardnerjr: `tabs` is missing from the values here, and it's a big one.  technically navigation is actually `nav`:
```
const enum ViewStyle {
    Bullets = "bullets",
    List = "list",
    Paragraph = "paragraph",
    Nav = "nav",
    Tabs = "tabs",
}
```

is there a thing for the actual allowed values when they are enums?

schema/workbook.json
```diff
@@ -1,2 +1,754 @@
...
+                    "const": 9
+                },
+                "content": {
+                    "type": "object",
+                    "properties": {
+                        "version": {
+                            "type": "string",
+                            "description": "Version of parameter step",
+                            "examples": [
+                                "KqlParameterItem/2.0"
```
gardnerjr: check on the versions, i think all of them except metrics should be default `whatever/1.0` ? instead of examples?

schema/workbook.json
```diff
@@ -1,2 +1,754 @@
...
+                            ]
+                        },
+                        "style": {
+                            "type": "string",
+                            "description": "style of parameter",
+                            "examples": [
+                                "pills"
```
gardnerjr: `pills|standard` here?

schema/workbook.json
```diff
@@ -1,2 +1,754 @@
...
+                "localizable"
+            ]
+        },
+        "chartLabelSettings": {
+            "$id": "#/definitions/chartLabelSettings",
+            "type": "object",
+            "properties": {
+                "seriesName": {
+                    "type": "string",
+                    "description": "The id of the property"
```
gardnerjr: nit: name of the series

schema/workbook.json
```diff
@@ -1,2 +1,754 @@
...
+                },
+                "label": {
+                    "type": "string",
+                    "description": "The custom label for this property",
```
gardnerjr: nit: label for this series

schema/workbook.json
```diff
@@ -1,2 +1,754 @@
...
+            "type": "object",
+            "properties": {
+                "columnId": {
+                    "type": "string",
+                    "description": "The id of the column"
+                },
+                "label": {
+                    "type": "string",
+                    "description": "The label of the column header",
+                    "tags": [
```
gardnerjr: here, do we haev any way of getting the comments to the localizers?  we tag these fields as localizable, but the comments here and in series labels aren't marked in any way so that the localizers see them?

schema/workbook.json
```diff
@@ -1,2 +1,754 @@
...
+                "linkIsContextBlade": {
+                    "type": "boolean",
+                    "description": "Open up the link in a context blade"
+                }
+            }
+        },
+        "parameterItem": {
+            "$id": "#/definitions/parameterItem",
+            "type": "object",
+            "properties": {
```
gardnerjr: i think parameter item also has a thing for dropdowns of what value to display for the "defaults" item that could be localized?

ericajlin: I wasn't able to find this option. If you are able to find it can you point me to it?

schema/workbook.json
ericajlin: From the email thread it doesn't seem like it, but we could ask

PR #473, opened 2019-10-28
cdi in azmon gallery

heyaWorld:


heyaWorld:
> do you need extra category info? where does this show up category wise when you add it to the gallery this way? does a cosmosdb category show up in the gallery?

It will show as "Azure Cosmos DB Insights" category in the gallery. 
"Azure Cosmos DB Insights" is already defined in the categoryResources.json

PR #478, opened 2019-10-30
Add Azure AD troubleshooting preview workbook: archived log date range

markwahl-msft:
In Azure monitor workbooks, where a user can pick a date range to query, the results may appear incomplete and missing some of what the user was expecting.   This can happen if the date range the user selects for starting is earlier than the data that's actually held in the Azure monitor database.  

E.g., in the Sign-in Analysis workbook, a user can select TimeRange of Custom, and put in their own date range - the date picker doesn't limit the user picking a start date they wish even if there is no data that was sent to Azure monitor from Azure AD during that time period.  

This can happen if 
 - the person using a workbook is an auditor/investigator who isn't aware of when the Azure AD integration was turned on,
 - the person using a workbook is different from the person who turned on the Azure AD integration,
 - the person who configured Azure AD integrations sent the audit and sign in logs to different workspaces, 
 - the person using a workbook doesn't realize that the workspace data retention cap is trimming out and deleting data they're looking for.   

In order that a user can know what is the likely date range, this workbook surfaces what data is held from Azure AD, the dates of the oldest and newest sign in and audit events.   While the integration might have been turned on earlier and simply no data was being sent for a period of time as nothing was occuring in Azure AD, this can at least prevent significant  discrepancies between what a user is expecting to find and what is stored.   If a user is wanting to look at, for instance, all sign ins to a particular app in 2019, the can check this view to find out how far back they would see data from.


markwahl-msft:
FYI @dhanyahk

Workbooks/Azure Active Directory TroubleShooting/ArchivedLogDateRange/ArchivedLogDateRange.workbook
```diff
@@ -0,0 +1,55 @@
...
+      "content": {
+        "json": "### Audit Log Date Range"
+      },
+      "name": "text - 3"
+    },
+    {
+      "type": 3,
+      "content": {
+        "version": "KqlItem/1.0",
+        "query": "let oldestAudit = AuditLogs | where TimeGenerated > ago(3653d) | order by TimeGenerated asc | limit 1 | project OldestAuditEvent=TimeGenerated,Type;\r\nAuditLogs | where TimeGenerated > ago(3653d) | order by TimeGenerated desc | limit 1 | project NewestAuditEvent=TimeGenerated,Type| join oldestAudit on Type | project OldestAuditEvent,NewestAuditEvent\r\n",
```
gardnerjr: is this the same thing but simpler? `AuditLogs | where TimeGenerated > ago(3653d) | summarize OldestAuditEvent=min(TimeGenerated), NewestAuditEvent=max(TimeGenerated) by Type`

markwahl-msft: Thanks for the input, that works.  Updated workbook.

PR #477, opened 2019-10-30
Updated explorer workbook

piyushanshu:

PR #479, opened 2019-10-30
fix build break

gardnerjr:
fix build break - how did this get through the CI build which runs the same tests?


gardnerjr:
i'm not sure how https://github.com/microsoft/Application-Insights-Workbooks/pull/477 got completed, given that the CI build failed for that PR and GitHub requires all the checks to pass.

PR #484, opened 2019-10-31
add tests to travis

vutran01:


gardnerjr:
what are you trying to do here? i don't think we want (or can) run our whole build in the travis thing. 
we just want whatever *checks* the full build is doing to be the same, so any "tests" that the full build does NEED to be in our travis build.  we don't want our travis build to BE the full build, as it will start downloading content from 17 github repos and merging them.  we don't want that at all, we just want the validations to be the same in both places.

right now that is mostly true because the full build runs the same tests that CI does.  but the full build has some other checks about files existing that does NOT exist in the CI build.  we want that.

vutran01:
remove building the templates and added more tests to check

test/jsonValidationTest.js
```diff
@@ -85,8 +97,46 @@ describe('Validating Workbooks...', () => {
...
+                    validateJsonExistForWorkbook(workbookPath, results, folder.substr(workbookPath.length+1))
+                });
+
+            done();
+        }, true, workbookPath);
+    });
 });
 
+function validateJsonExistForWorkbook(rootPath, results, file) {
+    let paths = getProgressivePaths(rootPath, file);
```
gardnerjr: the way this is written, doesn't this mean that if you have 10 workbooks in a folder, the parent folders get checked 10 times, since getProgressivePaths gives you every intermediate path?  probably fine but seems like it's a lot of extra checks as the tree gets bigger?

test/jsonValidationTest.js
```diff
@@ -42,6 +42,18 @@ describe('Validating Cohorts...', () => {
             done();
         });
     });
+    
+    it('Verifying cohort category or settings json exists', function (done) {
+        browseDirectory(cohortPath, (error, results) => {
+            if (error) throw error;
+            results.filter(file => file.substr(-9) === '.cohort')
+                .forEach(folder => {
+                    validateJsonExistForWorkbook(cohortPath, results, folder.substr(cohortPath.length+1))
```
gardnerjr: does this work for cohorts? don't the files end in `.cohort?`  getProgressivePaths only seems to be looking for `.workbook`

test/jsonValidationTest.js
vutran01: good catch. I did notice that as well, however, the test runs in 41ms for the entire tree so it's fast. usually the leaf if 3 folder deep and it was worth doing the extra check to match the entire folder name

test/jsonValidationTest.js
vutran01: should filter out cohort

PR #492, opened 2019-11-05
Wkahnza   patch 1 -  adding workbook server availability

WKahnZA:
Adding new workbook for server availability

availworkbook


Workbooks/Virtual Machines/Availability/server availability.workbook
```diff
@@ -0,0 +1,261 @@
...
+            "id": "bfe50469-7469-4c4d-a33e-0edbb43546c4",
+            "version": "KqlParameterItem/1.0",
+            "name": "Subscription",
+            "type": 6,
+            "isRequired": true,
+            "multiSelect": true,
+            "quote": "'",
+            "delimiter": ",",
+            "value": [
+              "value::all"
```
gardnerjr: you might want to look at some of the other "at scale" templates like storage/cosmos db that do queries to filter down subscriptions to only those with the resource type/etc you're looking for.  in this case, you're getting ALL subscriptions and using that by default.  which means for anyone with a lot of subs, the second step will immediately not work because more than 100 subs are selected.

the storage/cosmosdb examples query the appropriate subs, then autoselect the sub with the most of that kind of resource

Workbooks/Virtual Machines/Availability/server availability.workbook
```diff
@@ -0,0 +1,261 @@
...
+              ]
+            },
+            "timeContext": {
+              "durationMs": 0
+            },
+            "timeContextFromParameter": "TimeRange",
+            "queryType": 1,
+            "resourceType": "microsoft.resourcegraph/resources",
+            "value": [
+              "value::all"
```
gardnerjr: similarly here, this one defaults to "all" workspaces, in the selected subs, which will also fail if more than a specific number of workspaces are queried at once (200) or if those workspaces span too many regions (20).  While you may want to allow all to be selected, you still might want this to also just pick one by default?

Workbooks/Virtual Machines/Availability/server availability.workbook
```diff
@@ -0,0 +1,261 @@
...
+        "style": "pills",
+        "queryType": 0,
+        "resourceType": "microsoft.operationalinsights/workspaces"
+      },
+      "name": "parameters - 1"
+    },
+    {
+      "type": 1,
+      "content": {
+        "json": "### This Workbook calculates availability of VM's based on the specified time period."
```
gardnerjr: instead of using a text step as a header, probably should use this as the title of the grid below (in the query step's advanced settings), and could simplify to just "VM Availability"

Workbooks/Virtual Machines/Availability/tmp.tmp
```diff
@@ -0,0 +1 @@
+
```
gardnerjr: what's this tmp.tmp file?  can you delete this?

WKahnZA: i needed a file to create the folder, still new at this couldnt find a better way. I did delete it afterwards

Workbooks/Virtual Machines/Availability/server availability.workbook
```diff
@@ -0,0 +1,261 @@
...
+      },
+      "name": "text - 3"
+    },
+    {
+      "type": 3,
+      "content": {
+        "version": "KqlItem/1.0",
+        "query": "let timeRangeStart = {TimeRange:start};\r\nlet timeRangeEnd = {TimeRange:end};\r\nHeartbeat\r\n| where TimeGenerated > timeRangeStart and TimeGenerated < timeRangeEnd and  Computer in ({Servers})\r\n| summarize heartbeat_per_hour=count() by bin_at(TimeGenerated, 1h, timeRangeStart), Computer\r\n| extend available_per_hour=iff(heartbeat_per_hour>0, true, false)\r\n| summarize total_available_hours=countif(available_per_hour==true) by Computer \r\n| extend total_number_of_buckets=round((timeRangeEnd-timeRangeStart)/1h)\r\n| extend round(availability_rate=total_available_hours*100/total_number_of_buckets,2)\r\n| order by availability_rate asc",
```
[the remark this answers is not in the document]
WKahnZA: i tried to add the resource group to this the parameter but if it doesnt return just a server name then i can't use it in the query. My kql skills may be failing me

Workbooks/Virtual Machines/Availability/server availability.workbook
WKahnZA: updated and resubmitted

Workbooks/Virtual Machines/Availability/server availability.workbook
WKahnZA: updated and resubmitted

Workbooks/Virtual Machines/Availability/server availability.workbook
WKahnZA: updated and resubmitted

PR #499, opened 2019-11-13
add localization script to pipeline

vutran01:
- add script to clone localization repos during restore 
- remove de-de from language since repo is private and pipeline does not have access
- update pipeline build to generate json for each language

### DIFF
--- a/scripts/ProcessTemplates.ps1
+++ b/scripts/ProcessTemplates.ps1
@@ -14,7 +14,7 @@ $azureBlobFileNameBase = "community-templates-V2";
 $repoBaseName = "Application-Insights-Workbooks"
 $supportedLanguages = @(
     "cs-cz", 
-#    "de-de", Germany is a private repo and pipeline does not have access to clone
+    "de-de",
     $defaultLanguage, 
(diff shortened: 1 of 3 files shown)


scripts/ProcessTemplates.ps1
```diff
@@ -203,16 +207,22 @@ function Copy-Object {
...
+        }
         $repoName = "$repoBaseName.$lang"
         $repoPath = "$rootPath\$repoName"
         if (Test-Path $repoPath) {
-            Write-Host "Repo exist on disk, performing git pull ..."
-            Set-Location -Path $repoPath
-            git pull
+            Write-Host "Repo exist on disk, skipping $repoPath ..."
+            #Set-Location -Path $repoPath
+            #git pull
```
gardnerjr: isn't this whole section of testing the path and doing pull or clone not needed now? didn't the restore script at the top do all this work already?

vutran01: this would help when testing locally to clone and the pull can be added back if we can determine if it's running in the pipeline or not

PR #501, opened 2019-11-14
Volkanc sapmonitor fixforecast

Volkanco:
Fix forecasting views @tniek


gardnerjr:
this PR is more than a year old with no updates, closing without merging.

PR #507, opened 2019-11-21
Added grouping by resource group to Availability Workbook

WKahnZA:
Added grouping by resource group to Availability Workbook and removed old version with non capitalized name

image


WKahnZA:
updated incorrectly

PR #512, opened 2019-11-26
Improve AtScale Performance onboarding check

andrewki-msft:
Improve onboarding check to also check for `ServiceMapComputer_CL`. Creating dev version of the workbook for testing in PROD, then will promote to production version once verified.


nasundar:
This should not be shown when prompting onboarding message.

image

andrewki-msft:
Hide `Test` parameter when query fails

image

andrewki-msft:
> This should not be shown when prompting onboarding message.
> 
> image

Hide `Test` parameter when query fails

image

PR #520, opened 2019-12-02
SQL DB Resource Group Insights Fix

ericajlin:
Second and hopefully last fix for SQB DB resource group template. Previously I removed one metric due to failing metrics call. Turns out there are many restrictions on which metrics are enabled. 
https://docs.microsoft.com/en-us/azure/azure-monitor/platform/metrics-supported#microsoftsqlserversdatabases

Changing the template to only use metrics that are enabled for all sql types

### DIFF
--- a/Workbooks/Resource Groups/Sql databases/Sql databases.workbook
+++ b/Workbooks/Resource Groups/Sql databases/Sql databases.workbook
@@ -130,33 +130,19 @@
             "namespace": "microsoft.sql/servers/databases",
             "metric": "microsoft.sql/servers/databases-Basic-cpu_percent",
             "aggregation": 4,
-            "splitBy": null,
-            "splitBySortOrder": -1,
-            "splitByLimit": 5
-          },
-          {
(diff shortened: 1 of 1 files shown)


Workbooks/Resource Groups/Sql databases/Sql databases.workbook
```diff
@@ -216,48 +202,41 @@
...
-            {
-              "columnMatch": "microsoft.sql/servers/databases-Basic-log_write_percent Timeline",
+              "columnMatch": "microsoft.sql/servers/databases-Basic-connection_failed Timeline",
               "formatter": 5,
               "formatOptions": {
                 "showIcon": true
               }
             }
           ],
+          "rowLimit": 10000,
```
heyaWorld: do we need this line?

ericajlin: probably good to have a limit (even though it's very large)

PR #528, opened 2019-12-09
Wakahn branch3 - adding workbook for Azure resource locations

WKahnZA:
image

Need to confirm how to only show resources with a valid location (lat/lon)


WKahnZA:
Also please confirm settings.json is correct

Workbooks/Azure Resources/Azure Resource Locations/Azure Resource Locations.workbook
```diff
@@ -0,0 +1,95 @@
...
+          {
+            "id": "bfe50469-7469-4c4d-a33e-0edbb43546c4",
+            "version": "KqlParameterItem/1.0",
+            "name": "Subscription",
+            "type": 6,
+            "isRequired": true,
+            "multiSelect": true,
+            "quote": "'",
+            "delimiter": ",",
+            "query": "where type =~ 'microsoft.compute/virtualmachines'\r\n| summarize Count = count() by subscriptionId\r\n\t| order by Count desc\r\n\t| extend Rank = row_number()\r\n\t| project value = subscriptionId, label = subscriptionId, selected = Rank == 1",
```
gardnerjr: this query is only showing subs that have virtual machines.  intentional?

WKahnZA: Not intentional will update today

WKahnZA: Updated to include all subscriptions

WKahnZA:
Also improved colour scheme
image

image

Workbooks/Azure Resources/Azure Resource Locations/settings.json
```diff
@@ -0,0 +1,8 @@
+{
+    "$schema": "https://github.com/Microsoft/Application-Insights-Workbooks/blob/master/schema/settings.json",
+    "name":"Azure Resource Locations",
+    "author": "Microsoft",
+    "galleries": [
+        { "type": "workbook", "resourceType": "Azure Monitor", "categoryKey": "All", "order": 100  }
```
acearun: You already have a categoryResources.json at the lower level - you don't need a categoryKey. The category string (Azure Resources) will be picked from categoryResources.json.

WKahnZA: categorykey from settings.json

gardnerjr:
looks good, couple of questions:

1) why not allow all subs?  why a limit of 100?  azure resource graph allows up to 1000 subs, so setting the limit to 1000 and allowing all would probably let most people use "all" ?

2) similar question with resource types, why not allow multi select there and also support all?

WKahnZA:
Updated with recommendations above

WKahnZA:
reduced subscription limit to 100 until it can be raised

PR #541, opened 2019-12-13
Delete  files with upper-case 'PNG' as file extension

john-par:
Updates related to [User Story 1656771](https://mseng.visualstudio.com/TechnicalContent/_queries/edit/1656771).


john-par:
This PR is a minor change for a LOC fix.

I deleted seven files that were duplicates caused by [this commit](https://github.com/microsoft/Application-Insights-Workbooks/commit/ef03c40159623dcaedf04953ec18410dda0e1c3b). The file extensions were updated successfully, but GitHub held on to both versions of the file. That created an issue for the localization team. Description from the user story:

>  These images has similar name (difference is only in capitals), which cause blocking in OpenLoc. This issue prevents uploading files to iCMS, so we cannot localize. 

#sign-off

PR #547, opened 2020-01-08
Updating onboarding & key vault template w/ multi resource

austonli:
Updating the onboarding and insights template to use the resource scope picker for faster perf.
Removing redundant metrics from insights template and replacing old logs with enhanced logs


gardnerjr:
@austonli your PR descriptions should have some context that explains what it is, what's changing, why, etc.  screenshots of what's new/different are also helpful.  otherwise it's all just diffing json which isn't super easy to do

PR #561, opened 2020-01-17
C164292: Fixing indentation to have the content well formatted

v-maudel:
Localization team has reported source content issue that causes localized version to have broken/different format compared to en-us version.  
Description: Missing indentation is not letting the content to render properly

PR #563, opened 2020-01-20
C164290: Fixing indentation to have the content well formatted

v-maudel:
Localization team has reported source content issue that causes localized version to have broken/different format compared to en-us version.
Description: Missing indentation is not letting the content to render properly


v-maudel:
Hi @kumarvaranasi, can you please help us by approving and merging/sing in off (#sign-off-ces-pm) this PR as well?
This related to (https://github.com/microsoft/Application-Insights-Workbooks/pull/561) 
Regards

v-maudel:
Hi @kumarvaranasi, can you please help us by approving and merging/sing in off (#sign-off-ces-pm) this PR as well?
This related to (#561)
Regards

v-maudel:
Hi @kumarvaranasi, can you please help us by approving and merging/sing in off (#sign-off-ces-pm) this PR as well?
This related to (#561)
Regards

v-maudel:
Hi @kumarvaranasi, can you please help us by approving and merging/sing in off (#sign-off-ces-pm) this PR as well?
This related to (#561)
Regards

gardnerjr:
i'm concerned of why indentation is breaking loc?  what specific thing is wrong with the standard formatting that is being adjusted?

v-maudel:
> 
> 
> i'm concerned of why indentation is breaking loc? what specific thing is wrong with the standard formatting that is being adjusted?

Hi @gardnerjr, actually, the formatting and the displaying is not consistent, since this is a code that it is not showing as desired (It is broken). Please, note the first part I marked is plain text while the second is part of a code-block, it should be the same for both. 
image

v-maudel:
Hi @kumarvaranasi, @gardnerjr  carmonmills, can you please help by approving and merging this PR?
Regards

gardnerjr:
my concern is that all you've done is added 4 whitespace characters to the beginning of every line of this file, making the file 1400+ bytes longer for no reason.  the file is json, so i'm not sure why whitespace is important in the process.  if you need it to look correct in github, you should always be using the "Raw" option to see the raw, unformatted versions of files, as GitHub is just guessing at file types to show default formatting.

gardnerjr:
also, any changes you do to whitespace formatting here will just be overwritten if anyone ever updates the template again, so i need to know *exactly* what part of the process is broken here and how/why it is only happening in *some* templates?

v-maudel:
Hi @gardnerjr, I understand your point ans I think it is valid, we suggested the spaces addition to have it rendering "fine" in the normal view and to normalize the current two different displays (as I mentioned before) 
image

The issue occurs because in markdown, 4 spaces create a codeblock (like the \`\`\`). 
Please, take note that the original request states that the "normal" view in [Chinese versions](https://github.com/microsoftDocs/Application-Insights-Workbooks.zh-cn/blob/master/Workbooks/Intune%20Audit/Audit/IntuneAuditActivity.workbook) is broken and differs to the English one. 
Anyway, if this file is not meant to be seen with a "pretty" format from the "normal" Github view, I think I can let the team that reported the issue know

gardnerjr:
if teams are diffing files, they should be using real diff tools and looking at the raw versions of files, not looking at github's default formatted versions of files.  I'm looking for a way to force GitHub to treat *.workbook as json files, but haven't found an explicit way yet to force it to treat them as json
https://stackoverflow.com/questions/59954935/how-do-i-configure-default-view-of-files-in-github-to-treat-files-as-json

for now, i'd suggest closing all these PR's that are just adjusting the formatting of files, it is just adding extra bytes which will definitely show up as changes, and will just be overwritten the next time someone updates the workbook content.

v-maudel:
@gardnerjr , Actually, they receive the issues from a MS third party that checks consistency between en-us and translated files. When they find any mismatch, they report it to our team. (in this case the format breaks the conversion making all the locale content as plain text without formatting)
Also, I shared the comments with the production team and they asked if it can be fixed by removing the empty break-lines inside code (see screenshot) in order to avoid this "issue". (It was tested in a playground environment)
image

v-maudel:
@gardnerjr, can you please review our last suggestion?

gardnerjr:
if removing empty lines would help you, i'm not against that.  i'm not sure why these files would have empty lines in them, as things that normally generate them wouldn't add them, so i can only assume someone previously edited this file by hand and inserted them.

> (in this case the format breaks the conversion making all the locale content as plain text without formatting)

Again, though: These files are all JSON plain text.  spacing and formatting inside the file between elements (not inside strings) is unimportant.

My concern that this process is going to cause a LOT of these little problems where a few extra spaces that are unimportant in everything that read/writes these files will show up as important in the loc process and cause a lot of these hand editing cases.  if your changes only remove empty lines *i* wouldn't block you from checking in.  but the content in this repo isn't owned by me so every file like this you touch will require approval by some code owner that you'll have to explain this to.

v-maudel:
@gardnerjr Thanks for your understanding and also for the explanation. For current production process I made the changes here. https://github.com/microsoft/Application-Insights-Workbooks/pull/586

JFYI: I checked and according to production team, they don't receive many of this changes.

PR #565, opened 2020-01-22
Usage metrics workbooks for users, sessions and events

MSKaiLian:


Workbooks/UsageMetrics/Usage - Events/settings.json
```diff
@@ -0,0 +1,12 @@
+{
+    "$schema": "https://github.com/Microsoft/Application-Insights-Workbooks/blob/master/schema/settings.json",
+    "name": "Events",
+    "author": "Microsoft",
+    "galleries": [
+        {
+            "type": "usageMetrics",
```
gardnerjr: you're adding a new workbook type?  until someone also makes changes to enable this in the insightsportal extension, this should/would be showing you a warning in the portal that it is being ignored and using "workbook" instead.

Workbooks/UsageMetrics/Usage - Users/Usage - Users.workbook
```diff
@@ -0,0 +1,710 @@
...
+            "typeSettings": {
+              "additionalResourceOptions": []
+            },
+            "queryType": 0,
+            "resourceType": "microsoft.insights/components"
+          },
+          {
+            "id": "3774b890-2a45-46d8-9a92-01145cc1c4ba",
+            "version": "KqlParameterItem/1.0",
+            "name": "WithGrain",
```
gardnerjr: you might want to try converting params like this that just do a "case" to use the new "criteria" parameter, that way you don't run queries

https://github.com/microsoft/Application-Insights-Workbooks/blob/master/Documentation/Parameters/Criteria.md

Workbooks/UsageMetrics/Usage - Users/Usage - Users.workbook
```diff
@@ -0,0 +1,710 @@
...
+            "typeSettings": {
+              "additionalResourceOptions": []
+            },
+            "queryType": 0,
+            "resourceType": "microsoft.insights/components"
+          },
+          {
+            "id": "37b05d6c-ab7c-4190-94d8-4973fcf06c49",
+            "version": "KqlParameterItem/1.0",
+            "name": "TimeRange",
```
gardnerjr: why isn't this just using the standard TimeRange parameter type?  none of the values inside this query would get localized, and you wouldn't be able to select it in time range dropdowns? you'd only be able to use it in queries?

Workbooks/UsageMetrics/Usage - Users/Usage - Users.workbook
```diff
@@ -0,0 +1,710 @@
...
+            "typeSettings": {
+              "additionalResourceOptions": []
+            },
+            "jsonData": "[\"Time\", \"Property\"]"
+          },
+          {
+            "id": "8cd9003c-8e76-4d21-8b22-b7a11197f42f",
+            "version": "KqlParameterItem/1.0",
+            "name": "CountBy",
+            "label": "By value",
```
gardnerjr: you should also be setting the description/explanation on all these parameters so they have infoballoons in the workbook as well?

Workbooks/UsageMetrics/Usage - Users/Usage - Users.workbook
```diff
@@ -0,0 +1,710 @@
...
+            "id": "fd41f402-40ca-4dc5-bd29-38ec93f9d521",
+            "version": "KqlParameterItem/1.0",
+            "name": "UserEvent",
+            "label": "Selected events",
+            "type": 2,
+            "isRequired": true,
+            "multiSelect": true,
+            "quote": "'",
+            "delimiter": ",",
+            "query": "let timeRange = {TimeRange};\r\n   union {Whoused}\r\n    | where timestamp > ago(timeRange)\r\n    | project name, itemType\r\n    | order by itemType,name asc\r\n    | summarize test = count() by itemType,name\r\n    | project Id=name, Title=name, Selected=false\r\n    | union (\r\ndatatable(Id:string, Title:string, Selected:boolean)[\r\n'*', 'All', true\r\n])\r\n\r\n",
```
gardnerjr: instead of doing "*" this way inside the query text, you should use the options just below the query to enable "all" that way, with its value being * there:
image
when you do that, "all" is a special item that becomes unselected if other items are selected

Workbooks/UsageMetrics/Usage - Users/Usage - Users.workbook
```diff
@@ -0,0 +1,710 @@
...
..., 1==1, name in ({UserEvent}))\r\n    | extend Apdex=case(\r\n    duration <= threshold, \"Satisfied\",\r\n    duration <= 4*threshold, \"Tolerating\",\r\n    \"Frustrated\")\r\n    | summarize SESSIONS = dcount(session_Id), EVENTS = count() , PERFORMANCE=\"Null\", TotalSatisfied=sumif(itemCount, Apdex == \"Satisfied\"), TotalTolerating=sumif(itemCount, Apdex == \"Tolerating\"), TotalFrustrated=sumif(itemCount, Apdex == \"Frustrated\")\r\n    | project SESSIONS, EVENTS, ApdexScore = (TotalSatisfied + (TotalTolerating / 2)) / todouble(TotalSatisfied + TotalTolerating + TotalFrustrated)\r\n    | extend PERFORMANCE = case(\r\n    ApdexScore <= 0.5, \"Unacceptable\",\r\n    ApdexScore <= 0.7, \"Poor\",\r\n    ApdexScore <= 0.85, \"Fair\",\r\n    ApdexScore <= 0.94, \"Good\",\r\n    \"Excellent\")\r\n    | project SESSIONS, EVENTS, PERFORMANCE\r\n    | evaluate narrow()\r\n    | project Column, Value, id = case(Column =='SESSIONS', 'community-Workbooks%2FUsageMetrics%2FUsage%20-%20Sessions', Column =='EVENTS', 'community-Workbooks%2FUsageMetrics%2FUsage%20-%20Events','Community-Workbooks%2FPerformance%2FApdex'), Note = case(Column =='PERFORMANCE', 'Threshold 2000 milliseconds', '­\t')",
```
gardnerjr: for places returning counts, i'd set the number formatting to know that it is counts so it can be "nificied":
image
image

gardnerjr: this comment applies generally throughout all the grids showing counts, etc.  for the ones showing bars of counts, you also should set the minimum value to 0, otherwise it calculates bars between the largest and smallest value

MSKaiLian: is there a way to change font size? 
image

gardnerjr: you have the renderer set to bignumber, so that field will always be big.  you could leave the number there for the duration or whatever, and put that unnacceptable text into another column, and use that as the subtitle instead of the left part of the tile?

Workbooks/UsageMetrics/Usage - Events/settings.json
MSKaiLian: Yes, I will add a new workbook type.

Workbooks/UsageMetrics/Usage - Users/Usage - Users.workbook
MSKaiLian: I removed it because TimeRange changes to standard parameter type that includes a default grain.

Workbooks/UsageMetrics/Usage - Users/Usage - Users.workbook
MSKaiLian: changed

Workbooks/UsageMetrics/Usage - Users/Usage - Users.workbook
MSKaiLian: updated

CODEOWNERS
```diff
@@ -42,3 +42,5 @@
 /Workbooks/SapMonitor/** @tniek @microsoft/applicationinsights-devtools
 
 /Workbooks/Virtual*Machine*/** @andrewki-msft @nasundar
+
+/Workbooks/UsageMetrics/** @MSKaiLian
```
gardnerjr: you need to have at LEAST 2 people here, otherwise nobody will be able to approve your things, since you'd be the only owner and you can't approve your own things.  worst case, add our team (`@microsoft/applicationinsights-devtools`) as a secondary

Workbooks/UsageMetrics/Usage - Events/Usage - Events.workbook
```diff
@@ -0,0 +1,691 @@
...
...'Authenticated or anonymous traffic', 'Property',\r\n\t'client_Browser', 'Browser version', 'Property',\r\n\t'client_City', 'City', 'Property',\r\n\t'cloud_RoleInstance', 'Cloud role instance', 'Property',\r\n\t'cloud_RoleName', 'Cloud role name', 'Property',\r\n\t'client_CountryOrRegion', 'Country or region', 'Property',\r\n\t'client_Model', 'Device model', 'Property',\r\n\t'client_Type', 'Device type', 'Property',\r\n\t'operation_SyntheticSource', 'Is real user traffic', 'Property',\r\n    'name', 'Name', 'Property',\r\n\t'client_OS', 'Operating system', 'Property',\r\n\t'operation_Name', 'Operation name', 'Property',\r\n\t'performanceBucket', 'Performance', 'Property',\r\n\t'tostring(parseurl(url)[\"Host\"])', 'URL host', 'Property',\r\n\t'tostring(parseurl(url)[\"Path\"])', 'URL path', 'Property',\r\n\t'source', 'Request source', 'Property',\r\n\t'resultCode', 'Response code', 'Property',\r\n\t'operation_SyntheticSource', 'Source of synthetic traffic', 'Property',\r\n\t'client_CountryOrRegion', 'State or province', 'Property',\r\n\t'success', 'Successful request', 'Property'\r\n\t]), customDimension, timeTable \r\n\t| where Type == byType\r\n    | project Property, ColumnName",
```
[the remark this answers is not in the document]
MSKaiLian: Resolved

CODEOWNERS
MSKaiLian: add @microsoft/applicationinsights-devtools team

Workbooks/UsageMetrics/Usage - Users/Usage - Users.workbook
gardnerjr: i see you did this generally in tiles, but not in any grids?  they're all showing unformatted numbers?

image

image

PR #576, opened 2020-01-29
skip copy workbook in loc if exist

vutran01:
sometimes loc branches are missing some .json and .workbook files so the script copies them from master to loc branches when processing the templates. however, when the .workbook filename changes, this process could copy and have multiple .workbook in a directory and the tool will fail. this fix will check if .workbook already exist in the loc branches and skip copying another .workbook file over

### DIFF
--- a/scripts/ProcessTemplates.ps1
+++ b/scripts/ProcessTemplates.ps1
@@ -256,6 +256,14 @@ Function CopyFromEnuIfNotExist() {
         $fileName = $enuFile.Name
         $destinationFile = "$fullName\$fileName"
         if (!(Test-Path $destinationFile)) {
+            if ($fileName -like "*.workbook") {
+                $existingWorkbooks = Get-ChildItem -Path "$fullName\*" -Include *.workbook
+                if ($existingWorkbooks.Count -ne 0) {
+                    Write-Host ">>>>>> Skipping .workbook in $existingWorkbooks <<<<<<<<"
+                    continue
+                }    
+            }
+
             $fullPath = $enuFile.FullName
             Write-Host "[#WARNING: missing File]: copying file $fullPath to $fullName"
             # copy file from enu to localized folder


scripts/ProcessTemplates.ps1
```diff
@@ -256,6 +256,14 @@ Function CopyFromEnuIfNotExist() {
         $fileName = $enuFile.Name
         $destinationFile = "$fullName\$fileName"
         if (!(Test-Path $destinationFile)) {
+            if ($fileName -like "*.workbook") {
```
gardnerjr: comments would be good.  if I interpret this right, it is skipping files from the *localized* repo if the same file name doesn't exist in the English one?

so in this case, should the output show `>>>>> skipping...` and then just below it show `[#Warning: missing file]…` ?

gardnerjr: in the build I see this:
[INFO] [133] >>>>>>>>>>>>>>>>>> Skipping .workbook in S:\scripts\Application-Insights-Workbooks.cs-cz\Workbooks\View Designer\View Designer Transition Guide\Transition Guide.workbook <<<<<<<<

[INFO] [134] >>>>>>>>>>>>>>>>>> Skipping .workbook in S:\scripts\Application-Insights-Workbooks.cs-cz\Workbooks\View Designer\View Designer Transition Guide\Transition Guide.workbook <<<<<<<<
```
[INFO] [135] [#WARNING: missing File]: copying file S:\Workbooks\Virtual Machine Scale Sets - Network Dependencies\categoryResources.json to S:\scripts\Application-Insights-Workbooks.cs-cz\Workbooks\Virtual Machine Scale Sets - Network Dependencies

[INFO] [136] [#WARNING: missing File]: ...found S:\scripts\Application-Insights-Workbooks.cs-cz\Workbooks\Virtual Machine Scale Sets - Network Dependencies\categoryResources.json and replacing "en-us" with "cs-cz"
``` 

which implies to me that it skipped the file completely, does that mean that the English one won't be there in this language either?  or does some other process handle that?

vutran01: i will add more comments.

for the file that's missing in Loc compared with Enu, this checks that .workbook is the missing file and then get all .workbook files in the Loc folder to see if there's already a .workbook in that folder and skip processing. current folder ($fullname) is the Loc branch it's processing

vutran01: skipping means it detected another .workbook in the Loc branch so it won't copy over the enu version; this means the Loc branch will use an older .workbook version until the next time loc process sync the files from master. missing file indicates it detected missing files in Loc branch compared with enu and it's copying over that file before merging into a final json file

PR #583, opened 2020-02-04
Add schema to all categoryResource.json

ericajlin:
Add category resource schema to all category files, add validation check for schema


test/jsonValidationTest.js
```diff
@@ -201,7 +201,7 @@ function validateNoFromTemplateId(settings, file) {
 
 function validateCategory(category, file) {
```
vutran01: current test only checks existing category if it's correct, we'll need another to verify that it exist as well

ericajlin: I believe that already exists: See validateJsonExistForWorkbook

gardnerjr:
## Because of the huge amount of owners, we're going to bypass the CODEOWNERS for this change.  no functional changes are occuring, only adding a schema element to files that don't have them.  Just waiting for a travis ci outage to resolve before we merge this.

PR #588, opened 2020-02-04
ignore $schema

vutran01:
$schema is needed for localization on categoryResources.json, however, it's causing ALM Integration processing so crash. need to investigate but this should resolve this issue and let loc continue

### DIFF
--- a/scripts/ProcessTemplates.ps1
+++ b/scripts/ProcessTemplates.ps1
@@ -100,7 +100,8 @@ Function AddCategory() {
     param(
         [string] $categoryName,
         [Object] $categories,
-        [Object] $categorySettings
+        [Object] $categorySettings,
+        [string] $language
     )
 
     if ($categories.$categoryName) {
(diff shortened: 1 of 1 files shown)


scripts/ProcessTemplates.ps1
```diff
@@ -118,11 +118,14 @@ Function AddCategory() {
     $categories.$categoryName.DescriptionByLanguage = @{ }
 
     $categorySettings | Get-Member -type NoteProperty | Foreach-Object {
-        $languageProperties = $categorySettings.($_.name)
+        # ignore the $schema field which is needed for localization of categoryResources.json
+        if ($_.name -ne '$schema') {
```
gardnerjr: doesn't this mean that any extra properties in this file are going to cause this to fail?  should this be simplified to just grab fields it knows about instead of ignoring one specific field?

vutran01: the only expected fields here are the language field, we can change that

scripts/ProcessTemplates.ps1
```diff
@@ -118,8 +118,9 @@ Function AddCategory() {
     $categories.$categoryName.DescriptionByLanguage = @{ }
 
     $categorySettings | Get-Member -type NoteProperty | Foreach-Object {
-        # ignore the $schema field which is needed for localization of categoryResources.json
-        if ($_.name -ne '$schema') {
+        # only process languages field in categoryResources.json
```
gardnerjr: not being an expert at powershell, but doesn't this now loop over every property in the file and then n^2 check to see if it is in supported languages?  couldn't you just loop over all the supported languages and see if there's a property with that name, one loop instead of 2 nested ones?  it isn't *that* big of a deal since its just a couple properties and 17 languages, but still seems like extra work this way?

vutran01: since we know what the lang is we should just match that and eliminate the loops

PR #596, opened 2020-02-05
moving insights templates down, getting started up

yves-chan:


Workbooks/Azure Monitor - Getting Started/categoryResources.json
```diff
@@ -1,3 +1,3 @@
 {
-    "en-us": {"name":"Getting started with workbooks", "description": " ", "order": 2000}
+    "en-us": {"name":"Getting started with workbooks", "description": " ", "order": 150}
```
yves-chan: I based this number off of the Virtual Machines order which is 200

gardnerjr:
add a screenshot maybe of what the gallery looks like after?

yves-chan:
@gardnerjr yes I can do that. I assume there isn't a way to preview what the gallery would look like?

gardnerjr:
no, currently no good way to do that, since any change could affect *lots* of galleries.  maybe the CI build should output a giant tree view of what's where?  not sure there'd be an easy way to spot the differences though?

PR #608, opened 2020-02-13
Wakahn branch4 - New Workbook Azure Resource Health

WKahnZA:
New workbook for Azure Resource Health

image

image


gardnerjr:
i'd hold off on this one a bit, we have some fixes coming for some subtle issues in azure health (like sql database always getting back an "unknown")

WKahnZA:
Is there an ETA on that fix?

gardnerjr:
the fix for sql dbs is in today, will be through the pipeline over the next few days.  We have found that "master" databases are not supported (seemingly intended?)

WKahnZA:
I recreated the file, is this correct now?

gardnerjr:
actually it looks like maybe you already did that delete/rename and the problem was just in the *history* of the branch?

Workbooks/Azure Resources/Azure Resource Health/Azure Resource Health.workbook
```diff
@@ -0,0 +1,247 @@
...
+                "value::1",
+                "value::all"
+              ],
+              "includeAll": false
+            }
+          },
+          {
+            "id": "bdfc7b21-2dd7-4619-9f9e-7b81b698ad9c",
+            "version": "KqlParameterItem/1.0",
+            "name": "Resources",
```
gardnerjr: this parameter is called "Resources" but it really is resource types?  and i'm not sure how it works given that this gets used in the alerts resource type selector that way, but then the "subscriptions" parameter is passed as the resource parameter, so it doesn't seem to ever work for me?

you might want to change this from an arg query that returns all the resource types to querying ARM to get the resource types that resource health supports?

(if you swap out the queryType:1 and query lines in this param with:
```
            "query": "{\"version\":\"ARMEndpoint/1.0\",\"data\":null,\"headers\":[],\"method\":\"GET\",\"path\":\"/providers/Microsoft.ResourceHealth\",\"urlParams\":[{\"key\":\"api-version\",\"value\":\"2014-04-01\"},{\"key\":\"$expand\",\"value\":\"metadata\"}],\"batchDisabled\":false,\"transformers\":[{\"type\":\"jsonpath\",\"settings\":{\"tablePath\":\"$.metadata.supportedResourceTypes\",\"columns\":[]}}]}",
            "queryType": 12,
```
you'll get a resource type dropdown that only has the resource types that the health query provider supports

gardnerjr: not sure how/if this works?
you have it as a resource *type* parameter, but then below the subscriptions parameter is used?  doing it that way would cause the health provider to filter the subscriptions param to resources of the selected type?  so it should never work?

you could change this param to instead query arm for the health provider's supported resource types and show just the types that health works on? let me know if you need details how to do that.  not sure if that api returns "subscription" as an option or not?

WKahnZA: Please advise, I'm not sure how to do this correctly, I got a json error.

WKahnZA: Ok I think It's done

WKahnZA:
Updated resource picker to use ARM query and improved summary page by using tiles

gardnerjr:
@WKahnZA this PR is like 6 months old, is it still valid?  closing for now, you'll need to reopen and re-merge, etc.  a lot has changed since this PR was created.

PR #615, opened 2020-02-20
add json path parameter formatitng samples, add links to time format samples

gardnerjr:
Added documentation and samples for using jsonpath in parameter formatting


Documentation/Parameters/JSONPathParamSample.workbook
```diff
@@ -0,0 +1,32 @@
+{
```
ericajlin: do you want to link this in the parameters.md doc? I don't see this referenced anywhere

Documentation/Parameters/Parameters.md
```diff
@@ -56,11 +56,29 @@ This is how the workbook will look like in read-mode, in the "Pills" style.
...
+## Parameter formatting
 The _In Text_ section used the `label` of the parameter instead of its value. Parameters expose various such options depending on its type - e.g. time range pickers allow value, label, query, start, end and grain.
 
 Use the `Previews` section of the _Edit Parameter_ pane to see the expansion options for your parameter:
 
 ![Image showing a time range parameter options](../Images/Parameters-Time-Previews.png)
 
+For more specific examples of formatting times, see [Time formatting](./Time.md#Time-parameter-options)
 
+## Formatting parameters using JSONPath
```
gardnerjr: @ericajlin  the json path file is referenced here in the parameters doc

Documentation/Parameters/JSONPathParamSample.workbook
gardnerjr: oh, the workbook content itself is not linked, not sure if it needs to be.  this is more a best practice kind of thing that we should be doing that all of the screenshots/samples we use should have a sample workbook here in docs so someone can reproduce it.

Documentation/Parameters/Parameters.md
gardnerjr: nm, you meant the .workbook file, not the markdown sample.

PR #629, opened 2020-02-22
Activity Log - multi tenant example

gunnarcms:
Adding an example of a workbook to monitor activity log on a multi tenant environment. 

The report relies on Activity Log Diagnostics settings configured to send data to a Log Analytics workspace.

image

### DIFF
--- a/Workbooks/Lighthouse/Overview/Overview.workbook
+++ b/Workbooks/Lighthouse/Overview/Overview.workbook
@@ -1,0 +1,343 @@
+{
+  "version": "Notebook/1.0",
+  "items": [
+    {
+      "type": 1,
+      "content": {
+        "json": "## Lighthouse sample activity report\r\nDemo activity report\r\n<br/>\r\nVisit [View service provider activity](https://docs.microsoft.com/en-us/azure/lighthouse/how-to/view-service-provider-activity) for more detials on activity tracking\r\n<br/>\r\nVisit [Monitoring delegated resources at scale](https://docs.microsoft.com/en-us/azure/lighthouse/how-to/monitor-at-scale) for more information\r\n\r\n---"
+      },
+      "name": "text - 0"
+    },
+    {
+      "type": 9,
+      "content": {
+        "version": "KqlParameterItem/1.0",
+        "crossComponentResources": [
+          "value::all"
+        ],
+        "parameters": [
+          {
+            "id": "49c90f73-69bc-4f0b-b6be-515863bff46f",
+            "version": "KqlParameterItem/1.0",
+            "name": "Subscription",
+            "type": 6,
+            "isRequired": true,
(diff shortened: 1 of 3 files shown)


Workbooks/Lighthouse/Overview/Overview.workbook
```diff
@@ -0,0 +1,343 @@
...
+        "queryType": 1,
+        "resourceType": "microsoft.resourcegraph/resources"
+      },
+      "name": "parameters - 1"
+    },
+    {
+      "type": 3,
+      "content": {
+        "version": "KqlItem/1.0",
+        "query": "AzureActivity\r\n//| where Caller contains ({User})\r\n| where \"{User:lable}\" == \"All\" or Caller in ({User})\r\n| where Caller contains '@'\r\n| where ActivityStatusValue  == \"Success\" \r\n| where ResourceProviderValue <> \"\"\r\n| where \"{Resource:label}\" == \"All\" or ResourceProviderValue in ({Resource})\r\n| summarize Deletes = countif(OperationNameValue contains \"DELETE\"), Writes = countif(OperationNameValue contains \"WRITE\"),  Activities = (countif(OperationNameValue contains \"DELETE\") + countif(OperationNameValue contains \"WRITE\")) by bin_at(TimeGenerated, 1h, now())\r\n",
```
gardnerjr: `{user:lable}` is a mispelling?

Workbooks/Lighthouse/Overview/Overview.workbook
```diff
@@ -0,0 +1,343 @@
...
+              ],
+              "allowCustom": true
+            }
+          },
+          {
+            "id": "c757ee66-876f-4c53-b6bc-50c19fc9e227",
+            "version": "KqlParameterItem/1.0",
+            "name": "Resource",
+            "type": 5,
+            "isRequired": true,
```
gardnerjr: not sure what this resource parameter is doing.  it is set to be a "resource" (like azure resources), but appears to be doing a query to get resource *providers*, like "microsoft.compute"?  

If you intend to get resourece *types*, you can set this parameter to a `resource type` parameter instead of `resource` and, known resource types will get labels from the azure portal.

if you intend it to just be a generic dropdown, set it to `Dropdown` as the parameter type instead of `Resource`.  every value in a `Resource` parameter is run through code in the portal to parse out the resource name, and you don't need that here.

Workbooks/Lighthouse/Overview/Overview.workbook
gardnerjr: actually, your check here of `label == "all"` will only work in english, and will not work in other languages.

see https://github.com/microsoft/Application-Insights-Workbooks/blob/jgardner/missingTableHint/Documentation/Parameters/DropDown.md#special-casing-all for a better way to do this

gardnerjr:
@gunnarcms this PR is ~1 year old, is it still applicable? if so, update and reopen.

PR #631, opened 2020-02-22
Made changes to Security and Audit workbook to make it work in Gallery for both modes

rakshith210:
image


andrewki-msft:
Need to fix cases where user isn't onboarded properly for this workbook. I'm working on it.

andrewki-msft:
Submitted #635

PR #641, opened 2020-02-27
move individual storage templates to use groups

gardnerjr:
Before, the individual storage templates all have a links step at the top that emulates tabs, but opens child workbook blades:

image

each of those buttons is actually another template, with a very similar links step at the top but a different button marked primary.

After, the overview template has actual tabs, and all the tabs except the overview come from including those other templates:
image

the individual templates used in the other tabs no longer have the navigation UI at the top, they only do that one specific thing.

you can test this with https://portal.azure.com/?feature.workbookGalleryBranch=jgardner%2FmoveIndividualStorageToGroups
to redirect the gallery to this branch


Workbooks/Individual Storage/Overview/Overview.workbook
```diff
@@ -175,438 +160,160 @@
...
-          {
-            "cellValue": "https://aka.ms/saidoc",
-            "linkTarget": "Url",
-            "linkLabel": "Learn more",
-            "style": "link"
           }
         ]
       },
-      "name": "Navigation menu",
+      "name": "Navigation menu - Copy",
```
yves-chan: nit: delete "- copy"

PR #649, opened 2020-03-04
No subscription help message

yves-chan:
## PR Checklist

This is what it looks like with subs
hasSubs

Here it is in edit mode:
editMode

- The insights using this need to have a parameter called ResourceName to use render the resourceName in noSubscriptions workbook template


Workbooks/CosmosDb/Overview/Overview.workbook
```diff
@@ -153,1089 +149,1135 @@
...
       }
     },
     {
-      "type": 1,
+      "type": 12,
       "content": {
-        "json": "### Start editing the Overview section here"
+        "version": "NotebookGroup/1.0",
+        "groupType": "template",
+        "loadFromTemplateId": "/subscriptions/8980832b-9589-4ac2-b322-a6ae6a97f02b/resourceGroups/oychan-rg/providers/microsoft.insights/workbooks/4297e410-7054-4e23-bacc-5d8418698154",
```
yves-chan: TODO: change this to the community one upon check in

Workbooks/CosmosDb/Overview/Overview.workbook
```diff
@@ -153,1089 +149,1135 @@
...
-            "subTarget": "Failures",
-            "style": "link"
+            "id": "cf189bc4-17e3-4578-a115-5e544e064896",
+            "version": "KqlParameterItem/1.0",
+            "name": "hasSubscription",
+            "type": 1,
+            "value": "true",
+            "isHiddenWhenLocked": true,
+            "criteriaData": [
+              {
```
gardnerjr: why do you need this hasSubscription param at all?  you should be able to do this in the conditional visibility of just "subscriptions [equal to] [not set]" or "subscriptions [not equal to] [not set]" without an extra param

Workbooks/CosmosDb/Overview/Overview.workbook
gardnerjr: will be simpler to just check in the no subscriptions content one as its own single PR so you can reference it in this PR

yves-chan: Yeah will do that first, I'll make a new PR with this change

Workbooks/CosmosDb/Overview/Overview.workbook
yves-chan: that's just not inputting a value right? it will be 'not set'? I tried both ways but the conditional didn't quite seem to work.
image
The No Subscriptions group is still showing for some reason so I had to make a param

gardnerjr: because the name of the parameter is `Subscription`, not `Subscription_s_`:
image

yves-chan: 🤕 🤦‍♂ ​​​​​​​

yves-chan:
> tested via https://ms.portal.azure.com/?feature.workbookGalleryBranch=noSubscriptionHelpMessage and seems to be working?
> 
> the only thing i'm seeing is that while subs is being populated, the nosubscriptions thing is showing? that might be something we need to add general special case for in the visibility stuff to add "and is not loading" or something to parameter checks?

Yeah we can create a ticket for that to enhance the visibility options

PR #656, opened 2020-03-09
Update the build process for templates to create an NPM package of the templates.

gardnerjr:
## PR Checklist

Update the build process for templates to create an NPM package of the templates.

after this is done, will need to update the release pipeline to add a stage to grab the package output part and publish it to the msazure package repo that the InsightsPortal project uses to build. all of the template content will then be published via the hosting service and will be updated on the portal's standard schedule with the portal's hotfix process.

the template content will be treated the same way the portal's iframes are currently, where a specific package version is used in the deployment.

simplified *some* of the current script was change the way the localized repos were sync'd so that the whole language specific repo gets synced at once with en-us to make sure all the required content is there, instead of doing it piece by piece as things are enumerated

this build:
https://github-private.visualstudio.com/microsoft/_build/results?buildId=1423&view=results

has the new artifact:
image

which is a 9mb npm package, compared to the current language specific file artifact which is 18x16mb processed files deployed to storage and synced with alm service.

#### what's in that package?
```
package
│   package.json
│   
├───[each language]
│   └───Workbooks
│           Azure Monitor - Applications-Failure Analysis.json
│           ...about 157 other workbooks
│
│           _gallery.workbook-Azure Monitor.json
│           ...about 47 other galleries
│
│           _index.json
```
if you were to unpack the package, you'd find 4 kinds of things
1) the package.json describing the package, version, etc
    containing a folder for each language, containing
2) `_gallery.[type].[resource type].json` files which define the categories and all the templates in those categories in that gallery.  
3) individual workbook template files for that language
4) `_index.json`, a map of every template id to its file name.  this will allow us to support renaming/moving/replacing templates without breaking references/etc later.


gardnerjr:
although it looks like my changes that require msazure for the official build might have broken the CI build too, as now it needs auth to get that specific package (which the CI build does *not* need)

scripts/ProcessTemplates.ps1
```diff
@@ -401,6 +356,416 @@ Function BuildingTemplateJson() {
...
+
+
+        $fullGallery = @{}
+        $index = @{}
+        $allCategories = @{}
+        # find all the important files and run one pass for each kind
+        # 1) categoryresources.json - to create virtual galleries
+        # 2) settings.json, to find all the workbooks and what galleries they are in
+
+        $categories = Get-ChildItem "$currentPath\$reporttype" -Recurse -file -Include "categoryresources.json"
```
yves-chan: does casing matter? should it be categoryResources.json?

scripts/ProcessTemplates.ps1
```diff
@@ -401,6 +356,416 @@ Function BuildingTemplateJson() {
...
+
+
+    # find all the important files: **/categoryResources.json, *.workbook, *.cohort, **/settings.json, and any svg images
+    # in the source path, and make sure they exist in the specific language's path
+    foreach ($reportType in $reportTypes) {
+        $created = 0;
+        $reportPath = Join-Path -Path $sourcePath -ChildPath $reportType
+        $langPath = Join-Path -Path $sourcePath -ChildPath "scripts\$lang\$reporttype"
+        Write-Host "INFO: Syncing $lang with en-us from '$reportPath' to '$langPath'"
+        $files = Get-ChildItem $reportPath -Recurse -file -Include "categoryresources.json", "*.workbook", "*.cohort", "settings.json", "*.svg"
```
yves-chan: $categoryMetadataFileName

scripts/ProcessTemplates.ps1
```diff
@@ -342,19 +295,21 @@ Function BuildingTemplateJson() {
                 }
 
                 $categoryName = $category.Name
-
-                CopyFromEnuIfNotExist $category.FullName $language
                 $templates = Get-ChildItem $category.FullName
 
+    
                 $categorySettingsPath = Join-Path $category.FullName $categoryMetadataFileName 
+                if ($false -eq (Test-Path -Path $categorySettingsPath -PathType Leaf)) {
```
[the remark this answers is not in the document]
gardnerjr: strange.  this must have been left over from something i copied, as this was one of the things i fixed by redoing the way it sync'd data from en-us to the other languages

gardnerjr:
i've queued a build of this branch to make sure it will succeed before i actually complete the PR.

PR #658, opened 2020-03-11
Add VM Insights Menu workbook

andrewki-msft:
Also remove some unneeded templates. VM Insights workbook is a scope
selection for VM Insights workbooks.

Signed-off-by: andrewki-msft <43890980+andrewki-msft@users.noreply.github.com>

Tests
--

Context-free mode:
image

Hybrid mode, onboarded workspace:
image

Hybrid mode, non-onboarded workspace:
image
* `Workspaces` is unset because this workspace has no reporting machines, `Test` parameter will not have the value of `1` as a result and can be treated as a non-onboarded scenario by the consuming template.

Azure mode (subscription):
image

Azure mode (subscription > resource group with no machines):
image
* `Resource groups` is unset because this resource group has no machines. As a result, the `Test` parameter will not have the value of `1` and can be treated as a non-onboarded scenario by the consuming template.

Azure mode (subscription > resource group with machines):
image


gardnerjr:
@andrewki-msft this PR is almost a yea old, is it still necessary? if so, update and reopen, if not delete the branch?

PR #675, opened 2020-03-16
Adding storage-account-insights type

yves-chan:
Currently all the storage account workbooks are of type 'Workbook'. When switching to the WorkbookViewerBlades, they aren't recognized as insightsMode and the tool bar just becomes a refresh symbol. 

Looks like we need to create the 'storage-account-insights' workbook type and change all storage account workbook types to 'storage-account'. The development process will be similar to storage insights migration.

This PR is to copy the templates over to the storage account gallery with the type 'storage-account-insights'.

## PR Checklist


yves-chan:
Closed because we just needed to add storage-account-insights type from insights portal, no data changes are needed

PR #693, opened 2020-03-24
add tile size settings docs

gardnerjr:
add tile size settings docs and screenshots to tile settings markdown content
update workbook schema with tile settings, chart settings


gardnerjr:
### note: i'm waiting on the azure critical changes lockdown to be over before merging this.

PR #706, opened 2020-04-09
Adding myself as a CODEOWNER

erichrt:


erichrt:
Closing so that Devesh can add me as codeowner only for the LB subfolder

PR #716, opened 2020-04-15
add a validation check to verify there aren't multiple .workbook files in a directory

gardnerjr:
Adds a test to the CI build to verify that only one `.workbook` file exists in a directory


test/jsonValidationTest.js
```diff
@@ -203,19 +201,30 @@ function validateNoFromTemplateId(settings, file) {
     }
 }
 
+// validate that there's only one .workbook file in a folder, as only one is going to get picked up by the processing
+function validateSingleWorkbookFile(settings, file) {
+    let dir = path.dirname(file);
+    fs.readdir(dir, (err, list) => {
+        let workbooks = list.filter(s => s.endsWith(".workbook"));
+        if (workbooks.length > 1) {
```
yves-chan, line 209: does there have to be 1 file? if so maybe we check == 1?

gardnerjr: if there wasn't at least one file, you'd never get to this code, because the code got called for a given `.workbook` file. 

the problem is if there's _more_ than one, so i think leaving the code here like that makes it more readable this way.

PR #719, opened 2020-04-17
add validation checks for filenames with special characters

gardnerjr:
There were some templates with `&` in the filename or directory name that had various issues either being shown in the gallery or opening once selected.

added a test to fail PR's that have filenames with both `/` and `\` (happened once before) or any that have `#` or `?` or `&` in them.  could escape/encode those and possibly others but this will at least keep some of the obvious ones from failing in urls.


ericajlin:
Can you add a line in the Contributing.md about not adding special characters to the workbook name?

PR #726, opened 2020-04-22
add support for workbooks in deeply nested folders

gardnerjr:
Updates the template package processing to allow more deeply nested folders.  Several teams have run into this, you should now be able to make more deeply nested folder structures.  I haven't fully verified that the category names work out correctly though because of some other issues (like nobody really using this for real without other strange cases)

For network team, their 3 "missing" templates now show up in the index as workbook templates:
image


gardnerjr:
(however, it looks like they have PRs that will flatten out the hierarchy right now so nobody will technically be using it)

PR #731, opened 2020-04-23
Bugfixes for Key Vault Insights

austonli:
Fixed a bugged parameter in the at-resource workbook for Key Vault Insights.

Updated the gallery type for the at-resource workbook as well


gardnerjr:
you deleted a whole bunch of templates, you aren't loading them from inside groups?

PR #736, opened 2020-04-29
added parameter to select workspace

danielwood95:
## PR Checklist

I added a Workspace parameter so that customers with multiple workspaces in a given subscription can select which workspace they want to view data for. 

* [x] *if updating templates, it is helpful to post a screenshot of what the changes look like.*

image


gardnerjr:
in the template , you'd save with "All" as the selected value for that param, which daniel did:
```
            "value": [
              "value::all"
            ],
```

you can't pick a "specific" one to be default, because there's no way to know what the value would be before it runs.  if you don't want "all" to be the default, you just one "one", you'd have to add the "any one" option to the dropdown options, and then use that as the default value, and it would just pick the first one it finds.

danielwood95:
Kranthi would you prefer that I make the default "all" or "any"? There's no way for us to know which one to use if a customer has multiple workspaces tied to a subscription

KranthiKiranerusu:
Lets leave the default selection as "all". I am assuming that it lists only the workspaces that the user has access to.

PR #744, opened 2020-05-06
Final PR for Key Vault Public Preview

austonli:
- Fixing a few small typos and formatting changes
- Changed resource overview links to point to insights blade

### DIFF
--- a/Workbooks/KeyVault/AzureMonitor/AzureMonitor.workbook
+++ b/Workbooks/KeyVault/AzureMonitor/AzureMonitor.workbook
@@ -472,6 +472,7 @@
                     "formatter": 13,
                     "formatOptions": {
                       "linkTarget": "Resource",
+                      "subTarget": "insights",
                       "showIcon": true,
                       "workbookContext": {
                         "componentIdSource": "cell",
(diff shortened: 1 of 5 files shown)


Workbooks/KeyVault/AzureMonitor/AzureMonitor.workbook
```diff
@@ -472,6 +472,7 @@
                     "formatter": 13,
                     "formatOptions": {
                       "linkTarget": "Resource",
+                      "subTarget": "insights",
```
yves-chan, line 475: for now keyvault resources don't have insights right? we will add this later?

Workbooks/KeyVault/Overview/settings.json
```diff
@@ -2,5 +2,5 @@
     "$schema": "https://github.com/Microsoft/Application-Insights-Workbooks/blob/master/schema/settings.json",
     "name":"Key Vault Overview",
     "author": "Microsoft",    
-    "galleries": [{ "type": "keyvaults-insights", "resourceType": "microsoft.keyvault/vaults", "order": 100 }]
+    "galleries": [{ "type": "keyvaults-insights", "resourceType": "Microsoft.KeyVault/vaults", "order": 100 }]
```
yves-chan, line 5: should this be capitalized? I noticed in the current keyvault resource blade, when you click on the UsageNotebookBlade in debugger mode on mpac (ctrl+alt+d), in the console you can find that GalleryResourceType is undefined, 
image
so no templates will show. I think we will need to configure the menu item in keyvaults so we can assign a gallery type, like we do with storage accounts.

image

gardnerjr: if the gallery resource type isn't *explicitly* set, it comes from the resource type of componentId.  in the above case it would still be `microsoft.keyvault/vaults`
i'm not sure why "source" in your bottom screenshot is also a resource id, it shouldn't be.  it should be the source of the blade, not a resource id.  it's only used for telemetry, but having it be a resource means it has a huge cardinality.

yves-chan: I see, this was something I did for storage accounts, I'll submit a pr to fix the source

Workbooks/KeyVault/AzureMonitor/AzureMonitor.workbook
austonli: I think they do now? Or rather once their PR is deployed they will.

PR #749, opened 2020-05-07
update docs for open blade

yves-chan:
## PR Checklist

https://github.com/microsoft/Application-Insights-Workbooks/blob/oychan/openJsonBladeDocs/Documentation/Links/LinkActions.md#custom-view-link-settings


Documentation/Links/LinkActions.md
```diff
@@ -92,10 +92,22 @@ Use this to open Custom Views in the Azure Portal. Verify all of the configurati
...
+
+#### View Inputs
+There are two types of inputs, Grid and JSON. Use 'Grid' for simple key and value blade inputs or select 'JSON' to specify a nested json input.
+
+- Grid 
+    - `Parameter Name`: The name of the View input parameter
+    - `Parameter Comes From`: Where the value of the View parameter should come from. Select from `Cell`, `Column`, `Parameter`, or `Static Value`  in [Link sources](#link-sources). **NOTE** If `Static Value` is selected, the parameters can be resolved using brackets like `{paramName}` in the text box. Also, columns can be treated as parameters columns by appending `_column` after the column name like `{columnName_column}`.
```
gardnerjr, line 101: why did we decide to go with `{blah_column}` syntax for columns instead of something specific to columns, like `["name of column"]` like you'd do in kql?  does the braces thing work if the column names have spaces and special characters that params normally wouldn't allow?

yves-chan: I think we decided to go with {blah_column} so we could introduce formatting options like we did we params, the primary use case was to have a label formatter for resources, like {resource_column:label}. So since the param formatting was familiar we went with that. Just checked that columns with spaces won't be recognized a column param, so we'll have to fix that

gardnerjr: that's kindof why something specific to columns, like `["blah":specifier] that's *different* than params might be necessary so that the regexes/rules can stay the same as it is without breaking anything

yves-chan: right, I will log a bug to work on making columns it's own notation and then update these docs. I'll keep the _column formatting for backwards compat

PR #750, opened 2020-05-09
Update Node GPU.workbook

BLuEScioN:
## PR Checklist

* [x ] Explain your changes, so people looking at the PR know *what* and *why*, the code changes are the *how*.
The workbook queries fail when clusterId is present because the where clause does not put the clusterId in quotes 
* [x ] *if updating templates, it is helpful to post a screenshot of what the changes look like.*
No visual change

### DIFF
--- a/Workbooks/AKS/Node GPU/Node GPU.workbook
+++ b/Workbooks/AKS/Node GPU/Node GPU.workbook
@@ -15,7 +15,7 @@
             "description": "Filter data by time range",
             "isRequired": true,
             "value": {
-              "durationMs": 21600000
+              "durationMs": 14400000
             },
             "typeSettings": {
               "selectableValues": [
@@ -66,8 +66,7 @@
                 }
               ],
               "allowCustom": true
-            },
-            "resourceType": "microsoft.insights/components"
+            }
           },
           {
             "id": "bb8ed6dc-807e-4566-be76-26ac6ee7f90f",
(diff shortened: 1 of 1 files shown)


Workbooks/AKS/Node GPU/Node GPU.workbook
```diff
@@ -97,25 +98,27 @@
...
+            "resourceType": "microsoft.insights/components"
           },
           {
             "id": "51f0a81e-6eef-4129-b8a4-dc7dd4114aff",
             "version": "KqlParameterItem/1.0",
             "name": "clusterIdWhereClause",
             "type": 1,
             "description": "use this to filter by clusterid ",
-            "value": "| where \"a\" == \"a\"",
+            "value": "| where clusterId =~ \"/subscriptions/692aea0b-2d89-4e7e-ae30-fffe40782ee2/resourceGroups/nickb-rg-test/providers/Microsoft.ContainerService/managedClusters/nickb-aks-cluster-banner-test\"",
```
gardnerjr, line 111: this is going to fail to validate since the template has resource ids hardcoded in it

Workbooks/AKS/Node GPU/Node GPU.workbook
```diff
@@ -579,9 +583,9 @@
       "name": "links - 7"
     }
   ],
-  "fallbackResourceIds": [],
   "styleSettings": {
     "progressStyle": "loader"
   },
+  "fromTemplateId": "Community-Workbooks/AKS/Node GPU",
```
gardnerjr, line 589: this fromTemplateId line will also fail validation

Workbooks/AKS/Node GPU/Node GPU.workbook
BLuEScioN: Yea I loaded the workbook, made some edits, then copied the JSON...Because values are loaded into the workbook template, the editing experience is difficult and frustrating. Instead of going through the hassle of looking through the JSON and making ad hoc fixes the second time around, I just identified how my changes manifested in the JSON, then copied the JSON from github, and edited it according to the changes I saw in the JSON from my manual editing of the workbook...Then I loaded that JSON into the workbook to make sure I didn't mess things up. Vishwa brought up to Austin how it'd be great if there was an export button that stripped the values that were loaded into the workbook template, as well as the parameters that were changed while using the workbook. Obviously there is some nuance involved here, but hopefully you get the gist. I would love to see a feature like this in the future.

PR #756, opened 2020-05-12
update build pipelines to fix build issues after migration

gardnerjr:
update the build .yml files to use new build images, and update the test step to grab the test results as a TRX file instead of depending only on stderr output, which is always coming out now because the image has node.exe in multiple places.


gardnerjr:
interesting, looks like i also have to fix the CI build, i forget it gets used in different ways by travis and the real build

.pipelines/pipeline.user.windows.buddy.yml
```diff
@@ -57,8 +57,13 @@ test:
     - !!testcommand
       name: 'NPM Test'
       command: 'npmTest.cmd'
-      fail_on_stderr: true
+      fail_on_stderr: false
       skip_on_developer_box: true
+      testresults:                    # RECOMMENDED: If you want CDPX to collect any test results files and have the results be
```
yves-chan, line 62: whats CDPX stand for?

.pipelines/pipeline.user.windows.buddy.yml
```diff
@@ -5,7 +5,7 @@ environment:
     version: 2016
   runtime:
     provider: appcontainer
-    image: 'cdpxwin.azurecr.io/legacy/global/vse2017u5-external-azsdk-ext-win14393.1884:1.1-node10'
+    image: 'cdpxwin1809.azurecr.io/global/vse2019:latest'
```
yves-chan, line 8: What does changing this image do?

gardnerjr: it is having the build infra use a different machine image to run the build.  they required everything move to a version of windows on 1809+ for security updates, and then the rest of the line is stuff like what version of VS, etc.  in this case, workbook doesn't really need any of those dependencies since it's all just scripts.

.pipelines/pipeline.user.windows.buddy.yml
gardnerjr: [Cross-Platform Cloud Delivery Pipeline](https://onebranch.visualstudio.com/Pipeline/_wiki/wikis/Pipeline.wiki/330/OneBranch-Pipeline-General?anchor=%5Bthe-cross-platform-cloud-delivery-pipeline-(cdpx)%5D(%2Fcdpx-%252d-cross%252dplatform-cloud-delivery-pipeline))  its the internal org that does all the build/deployment architecture

PR #768, opened 2020-05-20
Modify performance Tile

MSKaiLian:
## PR Checklist

1. Remove empty table from who used
2. Fix query issue for getCustomDimension

### DIFF
--- a/Workbooks/UsageMetrics/Usage - Events/Usage - Events.workbook
+++ b/Workbooks/UsageMetrics/Usage - Events/Usage - Events.workbook
@@ -756,6 +756,7 @@
         "query": "{barChartQuery}",
         "size": 0,
         "aggregation": 3,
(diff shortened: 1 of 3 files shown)


Workbooks/UsageMetrics/Usage - Events/Usage - Events.workbook
```diff
@@ -756,6 +756,7 @@
         "query": "{barChartQuery}",
         "size": 0,
         "aggregation": 3,
+        "showAnnotations": true,
```
akkumari-ms, line 759: For my knowledge, what does this parameter do?

MSKaiLian: If the visualization shown has time as its x axis, setting this will show annotation markers at the top of the chart.

PR #773, opened 2020-05-25
Add chart type

MSKaiLian:
## PR Checklist

Add chart type selections(line, area) 
Update default value for 'Who used' 
Fix 'Requests,pageViews' in  'Who used' 
Show title when opening performance workbook
Remove Threshold number from performance tile
Set default value for durationThreshold
Set doNotRunWhenHidden for properties


Workbooks/UsageMetrics/Usage - Events/Usage - Events.workbook
```diff
@@ -1456,7 +1615,7 @@
           {
             "type": 1,
             "content": {
-              "json": "📄 Pageview ​ ‍ 📅 Event ​ ‍ 💬 Request"
+              "json": "������ Pageview ​ ‍ ������ Event ​ ‍ ������ Request"
```
OsvaldoRosado, line 1618: Did you mean to change the emoji? (Here and in other places below)

MSKaiLian: Thx. It looks like something wring. I will update it

PR #789, opened 2020-06-02
Users/nasundar/fixPinningTrend

nasundar:
1. Fix Top N network list for VMSS (Perf)
2. Fix query to fetch trend in top N lists for group of VMs and VMSS workbooks when pinned: In these workbooks, query snippet generation was done using a parameter (say {StandardQuery}) and that parameter was used to execute queries. In this case, when a user pins this list (containing trend), start and end time ranges were hardcoded as {TimeRange:start} would be substituted to an actual time in {StandardQuery} and the output of {StandardQuery} is saved in the pinned part. This causes empty trend part.

VMSS (InsightsMetrics):
image

image

VMSS (Perf):
image

image

At Scale (InsightsMetrics):
image

image

At Scale (Perf):
image

image


Workbooks/Virtual Machine Scale Sets - Performance Analysis/Performance Analysis (Perf)/Performance Analysis (Perf).workbook
```diff
@@ -360,36 +360,6 @@
...
...rceName = iff(NodeProps.type == 'AzureScaleSetNode', strcat(NodeProps.vmScaleSetName, ' | ', NodeProps.scaleSetInstanceId),          iff(NodeProps.type == 'AzureCloudServiceNode', strcat(NodeProps.cloudServiceRoleName, ' | ', NodeProps.cloudServiceInstanceId), Computer))     | project Computer, Type, ResourceId, ResourceName; let trend = Perf          | where TimeGenerated {TimeRange}         | where Computer in (computerList)          | where ObjectName == metric.object and CounterName == metric.counter        | make-series {TableTrend} default = 0 on TimeGenerated in range(datetime('\", timeRangeStart,\"'), datetime('\", timeRangeEnd ,\"'), totimespan('\", timeRangeGrain, \"')) by Computer     | project Computer, ['Trend ({TableTrend:label})'] = {TableTrend:label}; summaryPerComputer     | join kind=leftouter (trend) on Computer     | join kind=leftouter (NodeIdentityAndProps) on Computer     | join kind=leftouter (NodeIdentityAndPropsMin) on Computer     | join kind=leftouter (MachineSummary) on Computer     | project ResourceName, Type, {mergedAggregators}, ['Trend ({TableTrend:label})'], Properties = MachineSummary     | sort by {TableTrend:label} {tableTrendOrder}\"), ''));",
```
andrewki-msft, line 369: nit: I'd remove these query comments moving forward, but not a blocker

nasundar: Sure, I removed it initially, but thought you might have added it specifically because I did not see it other places/workbooks.

PR #790, opened 2020-06-02
Adding template for Azure Data Explorer (ADX) Clusters

cpaulsems:
image


yves-chan:
Looks good! A few things:
1. I would probably add the column formatting to use counts and the same number of fractional digits as our other insights where appropriate
2. in our existing insights, usually one of the columns are aggregated and the data is sorted using one of the columns. (I think there's a bug for aggregating percentages tho, so maybe if the tab only contains percentage data, we don't need this)

Workbooks/Insights/ADXCluster/Insights.workbook
```diff
@@ -0,0 +1,1441 @@
...
+        "value": ""
+      },
+      "name": "No Subscriptions group"
+    },
+    {
+      "type": 1,
+      "content": {
+        "json": "{Message}\r\n_Azure Data Explorer Cluster_\r\n<br />\r\n<br />"
+      },
+      "name": "text - 5",
```
yves-chan, line 182: can update this also

Workbooks/Insights/ADXCluster/Insights.workbook
```diff
@@ -0,0 +1,1441 @@
...
+          {
+            "cellValue": "selectedTab",
+            "linkTarget": "parameter",
+            "linkLabel": "Streaming Ingest Performance",
+            "subTarget": "Stream",
+            "style": "link"
+          }
+        ]
+      },
+      "name": "links - 2",
```
yves-chan, line 230: can update this

Workbooks/Insights/ADXCluster/Insights.workbook
```diff
@@ -0,0 +1,1466 @@
...
+                "{Resources}"
+              ],
+              "timeContext": {
+                "durationMs": 0
+              },
+              "timeContextFromParameter": "TimeRange",
+              "resourceType": "microsoft.kusto/clusters",
+              "resourceParameter": "Resources",
+              "metrics": [
+                {
```
yves-chan, line 834: if we move this metric object to the end of the metrics array, we can display it at the end of the columns

PR #795, opened 2020-06-03
Fix issue in WLI templates when VM and WS are in different subs

acearun:
Simple fix looks at all subs instead of just current sub.


Workbooks/Workloads/MySQL Insights/MySQL Insights.workbook
```diff
@@ -79,7 +79,7 @@
             "delimiter": ",",
             "query": " Resources\n| where id in~ ({Resources})\n| project ['Virtual machine'] = tolower(id), Name = name\n| join kind = inner (Resources\n    | where type =~ 'microsoft.compute/virtualmachines/extensions' and name =~ 'workload.linuxconfigagent'\n    | project Vm = extract(@'(.+)/extensions/.+', 1, tolower(id)), Insight = tostring(tags[\"monitoring-insight\"])\n    | where Vm != '' and Insight == '{WorkloadTag}'\n    ) on $left.['Virtual machine'] == $right.Vm\n| project Name\n",
             "crossComponentResources": [
-              "{Subscription}"
+              "value::all"
```
dougbrad, line 82: Shouldn't this be in just the Subscriptions they had picked?

acearun: > Good point. Fixed this and OnboardedComputerCount

PR #805, opened 2020-06-10
First draft of At Resource template for ADX Cluster

cpaulsems:
Overview page graphs are the same as for the existing
partner team resource blade. Export Performance, Ingestion Performance
and Query Performance metrics are the same as in At Scale.

image


Workbooks/Resource Insights/AzureDataExplorerCluster/settings.json
```diff
@@ -0,0 +1,7 @@
+{
+    "$schema": "https://github.com/Microsoft/Application-Insights-Workbooks/blob/master/schema/settings.json",
+    "name":"Azure Data Explorer Cluster Resource Overview",
+    "author": "Microsoft",
+    "isPreview": true,
+    "galleries": [{ "type": "workbook", "resourceType": "microsoft.kusto/clusters"}]
```
yves-chan, line 6: We might need to add the insights type also `{ "type": "adxcluster-insights", "resourceType": "microsoft.kusto/clusters", "order": 100 }`

PR #826, opened 2020-06-19
Update cohorts query

MSKaiLian:
## PR Checklist

* Update cohort data query


Workbooks/UsageMetrics/Usage - Events/Usage - Events.workbook
```diff
@@ -435,7 +435,7 @@
             "name": "cohortUsersQuery",
             "type": 1,
             "description": "Extract query in cohort json from UsersList dropdown ",
-            "query": "let query = '{UsersList}';\r\n print parse_json(query).cohort",
+            "query": "let query = '{UsersList:escape}';\r\nprint parse_json(query).cohort",
```
gardnerjr, line 438: this is just formatting json then taking a field out?  you can't just use criteria param and jsonpath?
isn't this just `{UsersList}:$.cohort` ?

Workbooks/UsageMetrics/Usage - Events/Usage - Events.workbook
```diff
@@ -450,7 +450,7 @@
             "name": "isBelongCohort",
             "type": 1,
             "description": "Extract IsBelong property in cohort json from UsersList dropdown ",
-            "query": "let query = '{UsersList}';\r\n print parse_json(query).isBelong",
+            "query": "let query = '{UsersList:escape}';\r\nprint parse_json(query).isBelong",
```
gardnerjr, line 453: same on all of these.  if you're just doing this to parse json, there are better ways that don't require running a query at all

Workbooks/UsageMetrics/Usage - Events/Usage - Events.workbook
```diff
@@ -630,7 +630,20 @@
...
               "durationMs": 0
             },
+            "timeContextFromParameter": "TimeRange"
+          },
+          {
+            "id": "d7a9bf96-471f-4a31-bd08-d7f999cae0f6",
+            "version": "KqlParameterItem/1.0",
+            "name": "parseCohortEvents",
+            "type": 1,
+            "query": "print replace(@'\\\\\"', @'\\\\\\\\\"', '{UserEvent:escape}')\r\n",
```
gardnerjr, line 640: `print replace(@'\\\\\"', @'\\\\\\\\\"', '{UserEvent:escape}')\r\n",`

this is very concerning to me.  you're using a kusto query to replace 4x quoting of a double quote with 8x quoting of a double quote?  this seems sooooooooo strange?

Workbooks/UsageMetrics/Usage - Sessions/Usage - Sessions.workbook
```diff
@@ -350,7 +350,7 @@
             "type": 1,
             "description": "Get bar Chart mode",
             "isRequired": true,
-            "query": "print iff('{UsageSplitBy}' == '<unset>' and '{isTimeRange}' == 'true', 'TimeBar', iff('{UsageSplitBy}' == '<unset>' and '{isTimeRange}' == 'false', 'ByValueBar', 'SplitBar'))",
+            "query": "let splitByValue = '{UsageSplitBy:escape}';\r\nprint iff(splitByValue == '<unset>' and '{isTimeRange}' == 'true', 'TimeBar', iff(splitByValue == '<unset>' and '{isTimeRange}' == 'false', 'ByValueBar', 'SplitBar'))",
```
gardnerjr, line 353: can you use criteria parameters for this instead of a query that just does print?  or no because of nested ands?

gardnerjr: actually 2 things: i think this won't work anywhere but english because `<unset>` would get localized

i think you can do this with criteria:
if {usagesplitby} is not empty then splitbar
else if {istimerange} is not empty then timebar
else byvaluebar

Workbooks/UsageMetrics/Usage - Events/Usage - Events.workbook
gardnerjr: (corrected my param syntax)

MSKaiLian: Thanks. I will update it.

Workbooks/UsageMetrics/Usage - Events/Usage - Events.workbook
MSKaiLian: same here. will update it.

Workbooks/UsageMetrics/Usage - Sessions/Usage - Sessions.workbook
MSKaiLian: <unset> is self defined value, not a label at here, won't localized. will consider to improve it at next pr.

PR #840, opened 2020-06-24
Fix wait stats tab for SQL insights

acearun:
Updates the queries to
1. Use just the wait_time_ms counter
2. Do delta calculations
3. Change units to milliseconds
4. Update column names

image


Workbooks/Workloads/SQL Azure Insights/SQL Azure Insights.workbook
```diff
@@ -2219,10 +2218,10 @@
...
..._ms'\r\n| extend Tags = todynamic(Tags)\r\n| extend Database = tostring(Tags[\"database_name'\"]), Instance = tostring(Tags.sql_instance), WaitType = tostring(Tags.wait_type)\r\n| order by Computer asc, Instance asc, Database asc, Namespace asc, Name asc, WaitType asc, TimeGenerated asc\r\n| extend Val = iff(Computer == prev(Computer) and Instance == prev(Instance) and Database == prev(Database) and Namespace == prev(Namespace) and Name == prev(Name) and WaitType == prev(WaitType) and Val > prev(Val), Val - prev(Val), real(null));\r\ndata\r\n| summarize Val = avg(Val) by WaitType, Database, Instance\r\n| extend Val = iff(isnan(Val), real(null), Val)\r\n| project Database = strcat(Instance, '/', Database), WaitType, Size = Val\r\n| extend jkey = strcat(Database, '/', WaitType)\r\n| join kind = inner (data\r\n    | make-series Trend = avg(Val) default = 0 on TimeGenerated from {TimeRange:start} to {TimeRange:end} step {TimeRange:grain} by WaitType, Instance, Database\r\n    | project Database = strcat(Instance, '/', Database), WaitType, Trend\r\n    | extend jkey = strcat(Database, '/', WaitType)\r\n    ) on jkey\r\n| project Database, WaitType, Size, ['Size trend'] = Trend\r\n\r\n",
```
dougbrad, line 2221: FYI - I did see bugs in some of other workbooks due to setting value to null instead of 0, just pointing out null not always correct if no changes, sometimes 0 is correct.

acearun: Thanks Doug - good point.

Zero value here is not correct. I believe NaN comes from there being only one entry for the wait stat in the query time range, causing us to now know what the delta is from the previous value. Setting it as null here allows us to represent the cell value as '-' instead of a value. I considered adding a '*' with a legend to indicate that the value changed but we don't know by how much. Choose this  approach to reduce cognitive load.

PR #842, opened 2020-06-25
Preview Insights Workbook path rename Part II

cpaulsems:
Second part of work started in https://github.com/microsoft/Application-Insights-Workbooks/pull/839. Copying EventHub, ServiceBus and EventGrid at scale to new directory structure.


cpaulsems:
> so be very careful moving any templates that are actively being used anywhere from any extension by their id

I should have copied the comments I made in the commits to the PR description, but yes the idea here is to make sure copies of the workbooks are left in the old location, and change history for the files are carried over to the new location.

yves-chan:
I see workbooks that are still in the 'Insights' folder which is good, we can delete those later after we change all the references for the templates on Az mon extension ToC

PR #852, opened 2020-07-02
Add hierarchy Azure SQL servers in SQLI

acearun:
image


Workbooks/Workloads/SQL Insights/SQL Insights.workbook
```diff
@@ -2875,5 +2889,8 @@
       "name": "Enable monitoring group"
     }
   ],
+  "fallbackResourceIds": [
+    "Azure Monitor"
```
dougbrad, line 2893: No impact of adding this?

acearun: Good catch - removed it.

PR #854, opened 2020-07-03
More cleanups to SQLI

acearun:
1. Name of server/db in the title of single db insights
2. Fix issue with db filter in wait stats
3. Add memory clerk filter in at-scale view.
4. Fix chart size of memory clerk in single-db view


acearun:
image

acearun:
Make it less colorful when healthy:

image

PR #894, opened 2020-07-30
removing categoryResource json - test

MGoedtel:
## PR Checklist

* [ ] *if adding new templates, ensure that your parameters and columns have display names.*

Removing redundant categoryResources.json file from the sub-folder under Azure Monitor - Workspaces to see if it resolves redundancy of category when viewing from AzMon\Workbooks.


MGoedtel:
- I need to package and upload to my storage blob to test and verify this change I'm testing works.  #please-hold this PR.

MGoedtel:
@yves-chan - I've adjusted strategy after consideration, aligning the parent folder for the workbooks with Azure Monitor agents instead of LA workspace.  This aligns with how it's positioned in the docs and with the service, including future plans around the agent.  I've tested this locally and it works so I'd like your approval to move forward.  As I add new workbooks/features, it makes it easy to keep contained in this "package". #sign-off

yves-chan:
Although this works I still think we should use[ virtual categories](https://github.com/microsoft/Application-Insights-Workbooks/blob/master/Documentation/Contributing.md#how-to-associate-any-existing-template-to-an-additional-category-in-workbooks)  to solve our folder structure issue
image
I could be wrong here, but theses workbooks don't necessarily need two different categories because the workbook themselves are related to workspaces.

Can we also set isPreview in the settings.json on the older workbook (or alternatively, we can just have an empty gallery array) to hide it from the gallery)?

yves-chan:
Had a quick sync with Matt on teams, in the future there would be a lot of different kinds of Agents, so it makes sense to start making agents it's own category. We've added a deprecated message in the old workbook.

We will add a workbook link inside the old workbook to point users to the new one. This will be done in a different PR

PR #904, opened 2020-08-13
[HOTFIX] SAP Monitor: Remove replay backlog fields from SapHana / HSR

tniek:
- To ensure compatibility with older HANA versions (< 2.0 SPS02), replay backlog fields were removed from the underlying SQL queries (see [PR 108 on payload side](https://github.com/Azure/AzureMonitorForSAPSolutions/pull/108)).
- This PR removes the pertinent fields (which are not ingested anymore) from the workbook templates as well. Otherwise, customers would see the following error message:
image

## PR Checklist

* [x] Explain your changes, so people looking at the PR know *what* and *why*, the code changes are the *how*.
* [ ] *if updating templates, it is helpful to post a screenshot of what the changes look like.*
* [x] *if updating templates, ensure that your parameters and steps have meaningful names.*
* [ ] *if adding new templates, ensure that your parameters and columns have display names.*
* [ ] *if adding new templates, or changing items in a gallery, it's helpful to post a screenshot of what the gallery looks like now*

PR #906, opened 2020-08-13
SAP Monitor: SapHana/Remove old workbook template

tniek:
- This PR removes the deprecated SAP HANA workbook template used in SAP Monitor < 2.0.

## PR Checklist

* [ ] *if adding new templates, ensure that your parameters and columns have display names.*


tniek:
> These are removed because we support backwards compatibility from this other PR #904, correct?

Actually this change is not related to PR #904.
The old HANA workbook template does not support multi-provider adequately, and we want to move customers over to 2.0. If the new (current) HANA workbook template is opened with an old (single provider) SAP Monitor, a message will be displayed to the user prompting them to upgrade (i.e. delete and re-create) to the current version.

PR #922, opened 2020-08-21
Fix usage calendar

acearun:
image


Workbooks/Usage/Usage Calendar/Usage Calendar.workbook
```diff
@@ -143,26 +124,54 @@
...
-        "aggregation": 0,
-        "showAnnotations": false,
-        "showAnalytics": false,
+        "queryType": 0,
         "resourceType": "microsoft.insights/components"
       },
-      "conditionalVisibility": null,
-      "customWidth": "35"
+      "customWidth": "45",
+      "name": "Usage calendar"
```
yves-chan, line 173: Looks like we need a `gridSettings` object to have persist labels (maybe this is a workbook bug?). but for now can you check the checkbox for 'Custom formatting` and reapply the labels? otherwise the second calendar looks like this

image

gardnerjr (changes requested):
the `Usage by hours of the day` template has the same problem, ordering times like

image

Workbooks/Usage/Usage Calendar/Usage Calendar.workbook
acearun: Addressed code review comments.

image

image

gardnerjr (approved):

PR #937, opened 2020-09-13
Azure Data Explorer Insights - Add "Commands and Queries" tab (log-based)

guregini:
Add "Commands and Queries" tab (log-based monitoring)

image


gardnerjr (commented):
is it intentional that there are some titles visible when the purple bar is visible:
image

also, is it necessary to have those titles as their own text steps instead of just having them be the title on the query/metrics themselves?  if the titles are separate, then when pinned, those items won't have titles either, and you have to manage visibility separately, etc.

guregini:
@gardnerjr, yes, it is intended that 2 titles are visible when the purple bar is visible, so users can see what 2 tiles/what section is based on the log-data. I can revalidate it with our UX expert, but as for now, this is the spec until the next round (will be soon).

Once there are logs, the tiles have a title. The "text steps" are there because the "real" title is longer. See image below:
image

As I said - I'll revalidate this issue for the next PR (we will edit some of the KQL queries for LA)

radennis (approved):

PR #990, opened 2020-10-22
Update Localization pipeline v2

ericajlin:
Localization pipeline v2 updates
- Add all files to pipeline
- Pull LCL files from localization repository
- Use identifiers for array-type localizable entries 

Note: This pipeline will not be used until existing templates are translated.


scripts/extractStrings.js
```diff
@@ -84,44 +122,95 @@ function openFile(file) {
...
+      }
       // Check for parameters that should be locked
-      findAndGenerateLockedStringComment(jsonKey, obj[i], outputMap);
+      findAndGenerateLockedStringComment(jsonKey, objectEntry, outputMap);
     }
   }
 }
 
+function canLocalize(text) {
+  const _notAllSpecialCharsRegex = new RegExp(".*[a-zA-Z].*", "g");
```
gardnerjr, line 156: i think you can do this as `new RegExp("[a-z]+", "i")` ?  anything a-z case insensitive anywhere, returning the first match only?  and i think you could create that const outside the function instead of creating a new one over and over?

yves-chan (approved):

PR #1010, opened 2020-10-29
Rename duplicate step names and parameter IDs

ericajlin:
Step names and parameter IDs help distinguish steps and parameters for localization and other workbook performance features. After a scan through existing templates, we have found the following templates with duplicate step names and GUIDS. Without this change, localization processes may be incomplete. This PR is to resolve duplicate IDs by adding a character to the end of the duplicates. Template authors will also be encouraged to revisit their workbooks to rename steps as necessary.


danielwood95 (approved):

andrewki-msft (approved):

gardnerjr (approved):

mpbastos (approved):
Changes are fine for "Windows Virtual Desktop" workbooks.

PR #1036, opened 2020-11-11
Becohen/add hd insight cluster level workbooks

blc16:
## PR Checklist

Adding resource level HDInsight workbooks. There are workbooks for each cluster type (Kafka, Hive, Hadoop, HBase, Spark, Storm). Also adding hdinsight-workbooks-team as owner of the HDInsight folder.

**Screenshots**
*Kafka Workbook*
DataflowMetrics
JVMMemory
Partitions
SystemMetricsDisk
ComponentLogs

*Hive Workbook*
ComponentLogs
Daemon
HiveJvm
HiveServer
HiveServerInteractive
YarnApps
YarnJVM
YarnNodeManagers
YarnOverview
YarnTimelineServer

*Hadoop Workbook*
Same as hive workbook above except for hive tab which is pictured directly below
HadoopHive

*HBase Workbook*
All sections besides HBase are the same as the above workbooks (YARN, System Metrics, Hive, and Component Logs are the other sections). Below are screenshots of the HBase section.
HBaseCompaction
HBaseHMasterRegionAssignment
HBaseJvm
HBaseMemstore
HBaseRegionServerOverview
HBaseRequestCounts
HBaseRequestLatency
HBaseWAL

*Spark Workbook*
SparkApps
SparkJobs
SparkOverview
SparkStages
SparkTasks

*Storm Workbook*
StormJvm
StormTopologyCapacity
StormTopologyLatency
StormTopologyMessaging
StormTopologyOverview


CODEOWNERS
```diff
@@ -30,7 +30,7 @@
 
 /Workbooks/Azure*Blockchain/** @keikumata @microsoft/applicationinsights-devtools
 
-/Workbooks/HDInsight/** @tylerfox @microsoft/applicationinsights-devtools
+/Workbooks/HDInsight/** @blc16 @hdinsight-workbooks-team @microsoft/applicationinsights-devtools
```
gardnerjr, line 33: this is intended to be the literal team, right?  `@microsoft/hdinsight-workbooks-team` is a valid team name, this is just a github user with that name, which is not valid?  and if you @blc16 are already in that team, you don't need to be mentioned explicitly here.

Workbooks/HDInsight/Workload-Specific/Kafka/settings.json
```diff
@@ -3,5 +3,5 @@
     "name":"HDInsight Kafka Insights",
     "author": "Microsoft",
     "isPreview": true,
-    "galleries": [{ "type": "workbook", "resourceType": "Azure Monitor", "order": 100 }]
```
gardnerjr, line 6: this change is removing this template from azure monitor all up, is that intentional?  it is possible for a template to be in both places, galleries is an array of entries.

gardnerjr (changes requested):
the codeowners part needs a fix for sure (invalid/nonexistant) users cause this file to also be invalid and breaks github's whole codeowners system (we have a support req already for this)

i didn't dig deep into the implementation of any of the workbooks, but did see that one of the screenshots has a step with a syntax error or other query error, so that's something to look at as well.

i also didn't dig to see how much is similar across the various templates, the groups feature allows you to "embed" the content of other templates into eachother, so if there is common stuff, it is good practice to extract that out to sub templates to simplify/share content between them.

CODEOWNERS
blc16: fixed this

Workbooks/HDInsight/Workload-Specific/Kafka/settings.json
blc16: We actually do not plan on adding a "Workbooks" ToC blade to our cluster portal just yet, so should I just revert this change and only show them in the Azure Monitor section?

gardnerjr (approved):

PR #1041, opened 2020-11-13
make PR validation fail if you have a settings.json file but no workbook (catch spelling/etc errors of filenames)

gardnerjr:
Add additional PR validation that catches case where people *think* they've added a workbook, but they haven't because of incorrect file names, like `something.workbook.txt` instead of `something.workbook` 

this PR will fail until the AAD team has fixed their workbook that has this problem right now.


test/jsonValidationTest.js
```diff
@@ -125,6 +127,16 @@ function validateJsonExistForWorkbook(rootPath, results, file) {
     });
 }
 
+function validateWorkbookExistForSettings(file) {
```
yves-chan, line 130: we also have a validateSingleWorkbookFile which checks if num of workbooks >1. Should we consolidate these checks for workbooks.length != 1?

gardnerjr: i don't think so.  because that rule basically runs **iff** it found a workbook in the folder.  not every folder *needs* to have workbooks, so there's no way to make sure this only works for 1 and only one, because then it would fail on other folders that don't have workbooks.

this one makes sure nobody puts more than one in a folder.

the new test makes sure if there's a settings.json, there's also exactly one workbook.

i could *probably* consolidate all these rules (there's a separate one for if you have a workbook it also finds settings.json or category resources, but at this point i'd like to have small simple obvious checks that might be redundant instead of a single big check that does all the things?

test/jsonValidationTest.js
```diff
@@ -125,6 +127,16 @@ function validateJsonExistForWorkbook(rootPath, results, file) {
...
+    let folder = path.dirname(file);
+    fs.readdir(folder, (err, list) => {
+        let workbooks = list.filter(s => s.endsWith(".workbook"));
```
vutran01, line 133: do we still support cohort? if so, you'll need to check for .cohort

yves-chan (approved):

test/jsonValidationTest.js
gardnerjr: true.  but nobody's touched cohorts in literal years, so i don't think its worth the time there.

gardnerjr:
wait, somehow this PR added  Submodule localization added at 3f93e0?  somehow the merge i did added this?
now i can't get rid of it?

PR #1047, opened 2020-11-16
Attempt to rename the nuget package tgz to avoid long path issues when releasing the package

gardnerjr:
Recently, there were some templates added that have a long enough folder/path name that the npm packaging steps of the release pipeline started failing.

It appears that the ADO extract task keeps the full filename and the nested folder's filename, creating a giant path like this:
`D:\a\r1\a\npm-publish\_azure_monitor_workbook_templates-1.0.0-build.15583.tgz_\azure_monitor_workbook_templates-1.0.0-build.15583.tar\package\en-us\workbooks...`

where the npm package+version number ends up in that path 2x, and then the names of the processed templates.  This PR tries to rename that build output to a shorter name (`build.tgz`) to try to get back 50-100 characters in the path. (i think it will only affect the first instance of the path, but i'm not sure, and without running the release pipeline on the output from this build, i won't know if it still gets packaged+correctly deployed with the right npm package name.

I've also opened up an issue with CDPX to see if there's some other settings for this i can turn on to avoid having to do this.

I've also notified the team that was affected by the long path.

Once i figure out what the resulting path is, i can add a validation to make sure folder paths don't exceed the limitations here.  (still sad/amazed that a 255 ch path limit is still a thing in 2020, but this *is* 2020 so it seems very apropos)

after the buddy build passed, i do see that the output is renamed:
image

but need to verify it on the real master build, and that the real deployment still finds the output and deploys the package to npm correctly.


scripts/renameOutput.cmd
```diff
@@ -0,0 +1,11 @@
+REM this will be a full name, like azure_monitor_workbook_templates-1.0.0-build.15583.tgz, and when the build/release try to unzip it, will use that full name in both the outer/inner folder names,
+REM so you end up with folder names for extracted files like D:\a\r1\a\npm-publish\_azure_monitor_workbook_templates-1.0.0-build.15583.tgz_\azure_monitor_workbook_templates-1.0.0-build.15583.tar\package\en-us\workbooks... which
+REM exceed various file/path operations for long template paths *inside* the workbook templates.
+REM for now, rename the file to build.tgz
+
+cd /D "%~dp0"
+cd "..\output\package"
+
+move azure_monitor_workbook_templates-* build.tgz
```
gardnerjr, line 9: this just renames whatever the generated file name was (will have a version# in it) to be always just build.tgz.
though maybe i should add `.tgz` to the end of the first filename to make sure it only renames that one file (though nothing else should be named that)

the actual versioning information is in package.json/.npmrc so the file name here shouldn't matter

vutran01 (approved):

scripts/renameOutput.cmd
gardnerjr: at least i think it is?  or its inside the package anyway?

gardnerjr:
i'm going to try to complete this, and see if the build AND deployment work. if they don't, i'll revert this

PR #1048, opened 2020-11-17
Add HelpContentSidePane workbook for ADXInsights

guregini:
Add HelpContentSidePane for ADXInsights


guregini:
settings.json was added in a different PR

gardnerjr (changes requested):
you can't do this and #1049 as different pull requests.
one will fail because you are trying to add a workbook with no settings..json metadata
the other one will fail because you're adding settings.json metadata with no workbook content

you need to do both of these things at the same time, in the same PR

guregini:
Closing this PR. The up to date PR is #1068

PR #1053, opened 2020-11-19
Localization v2 - Reduce output path under 260 char

ericajlin:
Windows has a max of 260 chars. Output path can be reduced by not including redundant template name as it is not necessary to include in the file path (folder used as unique identifier).

### DIFF
--- a/scripts/extractStrings.js
+++ b/scripts/extractStrings.js
@@ -103,7 +103,8 @@ function generateOutputPath(dir, folder) {
     outputPath = dir.replace(WorkbookTemplateFolder, folder);
   }
   if (!fs.existsSync(outputPath)) {
-    fs.mkdirSync(outputPath, { recursive: true });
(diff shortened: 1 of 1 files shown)


scripts/extractStrings.js
```diff
@@ -278,8 +278,8 @@ function generateLocProjectEntry(templatePath, resjsonOutputPath, rootDirectory)
 function getLocOutputPath(templatePath, extensionType, root) {
   const newTemplatePath = templatePath.replace(root, root.concat(LangOutputSpecifier))
   const templateSplit = newTemplatePath.split("\\");
```
gardnerjr, line 280: interesting, your stuff uses `\\` windows style paths?  the stuff in the tests using node is using `/` style unix paths? they both work locally and i the build infra?

gardnerjr (approved):

PR #1055, opened 2020-11-19
Add template path length validation to PR validation

gardnerjr:
* Turns off `--quiet` on npm pack in the build pipeline so we can track down if that's the thing truncating filenames or if it is some other part of the build before that

* adds a validation step to the PR validation that verifies that the path of the template doesn't exceed limits that appear to come from either the tgz format node uses, or from some limitation of the build pipeline


gardnerjr:
wait why does my branch show up as having all these other commits in it?  i'm going to redo this branch/PR

PR #1056, opened 2020-11-19
Add long file name/path validation to PR validation

gardnerjr:
Turns off --quiet on npm pack in the build pipeline so we can track down if that's the thing truncating filenames or if it is some other part of the build before that

adds a validation step to the PR validation that verifies that the path of the template doesn't exceed limits that appear to come from either the tgz format node uses, or from some limitation of the build pipeline


ericajlin (approved):

gardnerjr:
going to see how bad removing `--quiet` from npm pack is for the build, might put that back in later?

PR #1064, opened 2020-11-24
Minor changes in each query and renamed the workbook

shaaga:
This PR has following main changes:

1. Renamed the workbook 
2. Replaced summarize in the queries with make-series operator to account for the missing intervals
3. Added a warning incase no telemetry data is received


persiaAziz:
I am seeing no query results under the disk tab , first section. Any idea why that is?
osw

shaaga:
I am able to see the data now, not sure. Did you try refreshing ?
image

image

persiaAziz:
I just checked again. The current deployed os linux workbook works fine. But when I enter the changes you made, I see the same error

shaaga:
Its expected because we are filtering for only xfs and nfs4 type filesystems in the query and the vm on which the agent is running doesnt have any such fileysystems

persiaAziz (approved):

PR #1077, opened 2020-12-02
Add docs for upcoming time brush and global parameters features

gardnerjr:
## PR Checklist

Adding docs for time brush (both query and metrics) and info about global parameters to parameters docs.

none of these features is checked in or deployed to production environments but are available for testing via:
https://portal.azure.com/?feature.canmodifystamps=true&appInsightsExtension=workbooks


gardnerjr:
interesting. i turned on the new validation build settings, and it just shows as "expeted" but no way to make the check run? hopefully taking a merge from master will cause it to be queued?

gardnerjr:
Fixed up based on PR, also added new Limits.md to document various known datasource, visualization limits.

Documentation/DataSources/Limits.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Data Source Limits
+
+### General
+* In general, Workbooks limits the results of queries to be no more than 10,000 results. Any results after that point are truncated.
+* Each data source may have its own specific limits based on the limits of the service they query.
+* Those limits may be on the numbers of resources, regions, results returned, time ranges.  Consult the documentation for each service to find those limits.
+
+### Log based Queries (Log Analytics, Application Insights, resource centric queries)
+* Log Analytics [has limits](https://docs.microsoft.com/en-us/azure/azure-monitor/service-limits#log-queries-and-language) for the number of resources, workspaces, and regions involved in queries.
```
vutran01, line 12: don't we have a 250 limit on the query?

Documentation/Parameters/Parameters.md
```diff
@@ -37,20 +66,21 @@ This is how the workbook will look like in read-mode, in the "Pills" style.
...
 4. Run query to see the results
 
 ![Image showing a time range parameter referenced via bindings](../Images/Parameters-Time-Binding.png)
 
-### In KQL
-1. Add a query control to the workbook and select an Application Insights resource.
-2. In the KQL, enter a time scope filter using the parameter: `| where timestamp {TimeRange}`
+### In Logs/other queries
+1. Add a query control to the workbook and select the logs data source and an Application Insights resource or Log Analytics workspace.
+2. In the tet, enter a time scope filter using the parameter: `| where timestamp {TimeRange}`  (using the appropriate column name for your table)
```
vutran01, line 78: typo: "tet" -> "text"?

Documentation/Visualizations/Timebrush.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Brushing in a query chart
+
+When time brushing is enabled on a query chart, indicators will appear that the use can drag, or the user can "brush" a range on the time chart:
+
+![query time brush in progress](../Images/TimeBrush-query-brushing.png)
+
+Once the brush has stopped, the query chart will that range as a time range parameter, but will *not* zoom in (this is *different* than the behavior of metrics charts; Because of the complexity of user written queries, it may not be possible for workbooks to correctly update the range used by the query in the query content directly. If the query is using a time range parameter, it is possible to get this behavior by using a [global parameter](../Parameters/Parameters.md#Global-Parameters) instead)
+
+An icon in the toolbar in the upper right corner will also become active, to reset the time range back to its original, un-zoomed time range.
```
vutran01, line 31: i thought Manish added un-zoomed feature for metrics chart or like go back to previous view

vutran01 (approved):

Documentation/DataSources/Limits.md
gardnerjr: no, the grid limits rows that it shows, by default.  the hard query limit is 10k

Documentation/Visualizations/Timebrush.md
gardnerjr: yes, metrics only, by itself, when not using globals, because metrics knows one axis is always time and is itself in control of the queries going out.  but the logs charts do not, (if and only if the query was using a time range parameter would it be _possible_)

gardnerjr: (that's why this is explicitly called out here how query chart is different from metrics chart)

PR #1079, opened 2020-12-03
trying to add github action based pr build

gardnerjr:
I'm working on possibly switching from travis for our build to github actions, so this is in progress but trying to get it moving to see what happens :)


gardnerjr:
I think i have this action based build working.  using ubuntu and node 12, it takes ~25s.  with windows it takes ~1minute.
there's nothing platform specific about our tests so i think i'll pick the faster one :D.

Now i just have to figure out how to make github require it on PRs, which might require it to be completed and exist in the repo first?

vutran01 (approved):

PR #1087, opened 2020-12-05
David/smart insights

daweim0:
## PR Checklist

Adding a new workbook to test some "smart insights" features.

    - not sure how

### DIFF
--- a/Workbooks/AKS/Smart Insights/Smart Detectors.workbook
+++ b/Workbooks/AKS/Smart Insights/Smart Detectors.workbook
@@ -745,8 +745,5 @@
       }
     }
   ],
-  "fallbackResourceIds": [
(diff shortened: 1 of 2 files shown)


bragi92 (approved):

Workbooks/AKS/Smart Insights/Smart Detectors.workbook
```diff
@@ -61,7 +61,7 @@
             "crossComponentResources": [
               "value::selected"
             ],
-            "value": "/subscriptions/692aea0b-2d89-4e7e-ae30-fffe40782ee2/resourceGroups/jobyaks2/providers/Microsoft.OperationalInsights/workspaces/jobyaks2",
+            "value": "",
```
gardnerjr, line 64: you should remove the value field entirely.  leaving it like this may sets the value to "explicitly nothing selected", instead of letting the value be automatically set by a query, etc.

daweim0: Thanks for the suggestion, fixed

PR #1090, opened 2020-12-07
update github validation action to run on all prs

gardnerjr:
Updating the github action to run on all PRs, even between non-master branches, and to run whenever master is updated.
that isn't strictly necessary but the jarvis one did it too so it could have a link about successful status in the docs.


.github/workflows/template-validation.yaml
```diff
@@ -4,13 +4,9 @@
...
 
 on:
-  #push:
-  #  branches: [ $default-branch ]
   pull_request:
-    branches: [ master ]
-    # don't require the build on docs only changes?
-    #  paths-ignore:
-    #  - 'Documentation/**'
+  push:
```
yves-chan, line 8: does this mean validation is only run on a push? the pull_request field is empty

yves-chan: https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows nvm ,from the docs here it looks like this setting is correct

yves-chan (approved):

.github/workflows/template-validation.yaml
gardnerjr: yeah, i was confused too.  i think it is the yaml syntax, like "on push i have no extra settings, on pull request i do since they're embedded underneath.  without any settings i think you just do `on: [push, pullrequest]`

PR #1092, opened 2020-12-07
a PR to validate that pr's between non-master branches still get the validation build action run on them

gardnerjr:
a PR to validate that pr's between non-master branches still get the validation build action run on them
will validate and close without completing, and will delete this branch after testing is complete.


gardnerjr:
the PR did run the validation build, like it should have.

PR #1104, opened 2020-12-10
Deprecate the workload health datasource

gardnerjr:
## PR Checklist

The workload health (preview) data source is no longer functional, as the underlying service has stopped supporting the specific endpoints we were using.

Their suggested change is to use Azure Resource Graph (ARG) just like Alerts did with their resources.

Will need to get this updated in azure docs as well, as well as marking it actually deprecated in code in the extension like we did with alerts.


Documentation/DataSources/DataSources.md
```diff
@@ -10,13 +10,15 @@ Workbooks support these data sources:
...
-* [Workload Health](#workload-health)
 * [Azure Resource Health](#azure-resource-health)
 * [JSON](#json)
-* [Alerts](#alerts-preview)
 * [Custom Endpoint](#custom-endpoint)
 * Custom Provider
 
+Deprecated (Previously available):
+* [Alerts](#alerts-preview)
+* [Workload Health](#workload-health)
```
yves-chan, line 20: deprecated or removed?

gardnerjr: technically, both still exist in code. alerts works but can't be added to *new* workbooks, (works in ones where it already is used), workload health doesn't work and *will be* removed, but still also exists in code and will until it is removed and works through deployments.

yves-chan (approved):

PR #1126, opened 2020-12-16
remove cosmos sub picker

yves-chan:
- Remove cosmos sub picker

- Can test via oychan-cosmos scale and clicking on one of the cosmos db's in overview tab
image

## PR Checklist

### DIFF
--- a/Workbooks/CosmosDb/Resource Insights/Resource Insights.workbook
+++ b/Workbooks/CosmosDb/Resource Insights/Resource Insights.workbook
@@ -7,49 +7,19 @@
         "version": "KqlParameterItem/1.0",
         "parameters": [
           {
-            "id": "1a5523c4-8b8f-405b-915b-29f2bb09ad6b",
-            "version": "KqlParameterItem/1.0",
-            "name": "Subscription",
-            "label": "Subscriptions",
-            "type": 6,
-            "description": "All subscriptions with CosmosDB accounts",
-            "isRequired": true,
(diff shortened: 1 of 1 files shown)


Workbooks/CosmosDb/Resource Insights/Resource Insights.workbook
```diff
@@ -6,50 +6,20 @@
...
-            "timeContext": {
-              "durationMs": 86400000
-            },
-            "queryType": 1,
-            "resourceType": "microsoft.resourcegraph/resources"
-          },
           {
             "id": "aa83f370-5c47-4eb5-b44a-fedcad97f71b",
             "version": "KqlParameterItem/1.0",
             "name": "Resources",
```
gardnerjr, line 12: so this is intended to be an "at resource" template, but is powered by being fed a "Resources" parameter.  this isn't how at resource templates usually work, is it?  this means when they write code to open this blade, they'll also have to pass this as a parameter in the NotebookParams field when they open the workbooks blade with this template?  this will NOT work if this item appears in a gallery, nothign will ever set the param!

gardnerjr: yves showed that this IS working and that in his example he is passing the resource as component id.

gardnerjr (approved):
as i asked before, has anyone actually tested that the gallery settings in the settings.json are correct?  if they are not, this template won't be correctable and appear in a gallery until the freeze expires in January

Workbooks/CosmosDb/Resource Insights/Resource Insights.workbook
yves-chan: for now this is on all resource templates so I thought it was standard. Clicking on the template from the gallery opens the workbook with the right resource

yves-chan:
yes just tested, that the settings.json are correct for opening this template from the cosmos gallery

PR #1163, opened 2021-01-08
On behalf of Carter to add this downtime & outages workbook

MSKaiLian:
## PR Checklist

* On behalf of Carter to add this downtime & outages workbook provide easy **SLA** reporting & outage information for services


gardnerjr (changes requested):
it looks like the workbook is using "All" app insights resources by default?  you'll probalby want to not do that for just 3 reasons off the top of my head:
1) "all" could be > 100 and fail at draft because of limits of # resources
2) "all" could be across more than N regions, which will fail because it has a limit there
3) "all" could mix between AI and workspace based AI instances, and i think queries fail if you cross between those right now? i see this when i use the workbook:
image

similarly, you have the default time range to 30 days, which is going to query across cold storage as well, making this way slower if not timing out entirely.  

i wouldn't hide the time range or the resources inside groups like this, when things fail people will have to expand things until they figure out why

casocha:
@gardnerjr requested changes addressed in most recent file

gardnerjr (approved):
there are some strange quirks about how much space the params and explanations take up, but i don't see any "blocking" things

PR #1182, opened 2021-01-17
Add a new workbook template for Intune Devices specialist report

panqingbao:
Add a new workbook category for Intune Devices, which contains one new workbook template for Intune Devices specialist report. 

IntuneDevicesWorkbookTemplate_1
IntuneDevicesWorkbookTemplate_2


Workbooks/Intune Devices/Devices/IntuneDevices.workbook
```diff
@@ -0,0 +1,351 @@
...
+        "style": "pills",
+        "queryType": 0,
+        "resourceType": "microsoft.operationalinsights/workspaces"
+      },
+      "name": "Parameters"
+    },
+    {
+      "type": 1,
+      "content": {
+        "json": "## Number Of Devices Trends\r\n---"
```
gardnerjr, line 74: suggestion: instead of using all these extra text items as headers, i'd suggest using the header field in advanced settings for the queries themselves, for a few (mostly minor) reasons:
1. keeps the headers similar to all the other headers in other places in the portal
2. makes the workbook simpler + smaller, as it doesn't have all the extra text content
3. headers set directly in query/metrics steps are also used as the aria-label on those charts/grids for accessibility.  otherwise, they end up with the names of the step (which you've filled in so that's good)

Workbooks/Intune Devices/Devices/IntuneDevices.workbook
```diff
@@ -0,0 +1,351 @@
...
+        "json": "## Devices By Ownership\r\n---"
+      },
+      "customWidth": "50",
+      "name": "Text - DevicesByOwnership"
+    },
+    {
+      "type": 3,
+      "content": {
+        "version": "KqlItem/1.0",
+        "query": "IntuneDevices\r\n| where TimeGenerated > ago(1d)\r\n| where Result != \"Completed\"\r\n| join kind = inner (\r\n    IntuneDevices\r\n    | where TimeGenerated > ago(1d)\r\n    | where Result == \"Completed\"\r\n    | summarize arg_max(TimeGenerated, BatchId)\r\n) on BatchId\r\n| summarize arg_max(DeviceId, *) by DeviceId\r\n| summarize count() by OS",
```
gardnerjr, line 116: you have a time range parameter but don't appear to be using it?  it looks like the queries are all hardcoded to ago(1d)?

it looks like you could just search/replace `where TimeGenerated > ago(1d)` with `where TimeGenerated {TimeRange}` ?

Workbooks/Intune Devices/Devices/IntuneDevices.workbook
```diff
@@ -0,0 +1,351 @@
...
+      "name": "Text - iOSDeviceCountByVersion",
+      "styleSettings": {
+        "maxWidth": "50"
+      }
+    },
+    {
+      "type": 3,
+      "content": {
+        "version": "KqlItem/1.0",
+        "query": "IntuneDevices\r\n| where TimeGenerated > ago(1d)\r\n| where Result != \"Completed\"\r\n| where OS contains \"Android\"\r\n| join kind = inner (\r\n    IntuneDevices\r\n    | where TimeGenerated > ago(1d)\r\n    | where Result == \"Completed\"\r\n    | summarize arg_max(TimeGenerated, BatchId)\r\n) on BatchId\r\n| summarize arg_max(DeviceId, *) by DeviceId\r\n| summarize DeviceCount = count() by OSVersion\r\n| sort by OSVersion desc\r\n| project AndroidVersion = OSVersion, DeviceCount",
```
gardnerjr, line 266: not an expert on your data or schema, so just a question: can you use `has` in some of the places you are using `contains` ?  `has` uses an index and can be a lot faster than `contains`, if it gets you the right answer

Workbooks/Intune Devices/Devices/IntuneDevices.workbook
```diff
@@ -0,0 +1,351 @@
...
+      "content": {
+        "version": "KqlItem/1.0",
+        "query": "IntuneDevices\r\n| where Result != \"Completed\"\r\n| summarize NumberOfDevices = dcount(DeviceId) by BatchId\r\n| join (\r\nIntuneDevices\r\n| where Result == \"Completed\"\r\n| project BatchTime = TimeGenerated, BatchId\r\n| summarize arg_max(BatchId, *) by BatchId\r\n) on BatchId\r\n| project BatchTime, NumberOfDevices",
+        "size": 3,
+        "aggregation": 5,
+        "showAnnotations": true,
+        "timeContext": {
+          "durationMs": 0
+        },
+        "timeContextFromParameter": "TimeRange",
```
gardnerjr, line 89: this query seems to be the only one using the time range parameter?

panqingbao: Yes, we only use Time Range for trends query. The other queries will based on the latest daily snapshot (> ago(1day).

Workbooks/Intune Devices/Devices/IntuneDevices.workbook
panqingbao: Good point. Fixed!

Workbooks/Intune Devices/Devices/IntuneDevices.workbook
panqingbao: Make sense. Text items are removed and Header title is set instead.

Workbooks/Intune Devices/Devices/IntuneDevices.workbook
gardnerjr: it's kindof a strange thing though, that one shows me what i picked, and the others don't.  i'm not the PM of your feature but it is a little strange.

Workbooks/Intune Devices/Devices/IntuneDevices.workbook
```diff
@@ -0,0 +1,270 @@
...
+      "name": "Query - AndroidDeviceCountByVersion",
+      "styleSettings": {
+        "maxWidth": "50"
+      }
+    },
+    {
+      "type": 3,
+      "content": {
+        "version": "KqlItem/1.0",
+        "query": "IntuneDevices\r\n| where TimeGenerated > ago(1d)\r\n| where Result != \"Completed\"\r\n| where OS has \"iOS\"\r\n| join kind = inner (\r\n    IntuneDevices\r\n    | where TimeGenerated > ago(1d)\r\n    | where Result == \"Completed\"\r\n    | summarize arg_max(TimeGenerated, BatchId)\r\n) on BatchId\r\n| summarize arg_max(DeviceId, *) by DeviceId\r\n| summarize DeviceCount = count() by OSVersion\r\n| sort by OSVersion desc\r\n| project AndroidVersion = OSVersion, DeviceCount",
```
gardnerjr, line 220: FYI, all these other types have "AndroidVersion" as the column name, even for iOS and windows

gardnerjr (approved):
approved, but you have "android" in a couple column names, and the time range only applying to one chart instead of all the charts is odd

Workbooks/Intune Devices/Devices/IntuneDevices.workbook
panqingbao: Good catch! That is not correct. I will fix it.

panqingbao: Fixed!

PR #1214, opened 2021-02-03
Downtime

cartersocha:
error message updates & bug in the web test parameter where App Insights as the log choice wasn't being selected.

### DIFF
--- a/Workbooks/Availability/Downtime Outages/Downtime Outages.workbook
+++ b/Workbooks/Availability/Downtime Outages/Downtime Outages.workbook
@@ -1,6 +1,5 @@
 {
   "version": "Notebook/1.0",
-  "isLocked": true,
   "items": [
     {
       "type": 11,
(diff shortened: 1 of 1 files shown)


Workbooks/Availability/Downtime Outages/Downtime Outages.workbook
```diff
@@ -1450,14 +1441,20 @@
...
+                  "fieldName": "location",
+                  "parameterName": "mapLocations",
+                  "parameterType": 8
+                }
+              ],
               "queryType": 0,
               "resourceType": "microsoft.insights/components",
-              "crossComponentResources": [
-                "{Resources}"
-              ],
```
gardnerjr, line 1460: abunch of places here have removed the selected parameter which is going to break them. not sure what/how this came in but isn't good

gardnerjr (changes requested):

Workbooks/Availability/Downtime Outages/Downtime Outages.workbook
cartersocha: Added back to match Kais original version

gardnerjr (approved):
my only comments now are about step names, ideally they'd all have useful+distinct names, not be things like `query 9 - copy`, etc.  especially helpful for loc later, etc so if a step moves around the loc moves properly with it.

PR #1278, opened 2021-02-23
Template Localization V2

ericajlin:
- Switch localization from MicrosoftDocs to Azure localization pipeline
- Add test for duplicate ID
- Fix various templates to remove duplicate IDs


.pipelines/pipeline.user.windows.buddy.yml
```diff
@@ -70,7 +66,15 @@ build:
...
-            - '**/*' 
+            - '**/*'
+    - !!buildcommand
+      name: "Generate Templates"
+      command: 'scripts\generateTemplates.cmd'
+      artifacts:
+        - from: 'out\loc'
+          to: 'loc'
+          include:
+            - '**/*'
```
gardnerjr, line 77: i can never remember the rules here.  do *both* of these steps need to dclare they have output artifacts?  or can the first one be just a `!!defaultcommand`, and the second one be a `!!buildcommand` with outputs? you don't get 2 different copies of the outputs this way?

scripts/extractStrings.js
```diff
@@ -167,14 +205,16 @@ function getLocalizeableStrings(obj, key, outputMap) {
...
+  return !isNaN(parseFloat(n)) && isFinite(n);
+}
+
+/** Returns true if string has localizeable text  */
 function canLocalize(text) {
-  if (text != null && text.match(NotAllSpecialCharsRegex)) {
-    return true;
-  }
-  return false;
+  return text !== null && text !== "" && text.match(NotAllSpecialCharsRegex);
```
gardnerjr, line 214: >text !== null [](start = 9, length = 13)

given javascript shenanigans, check for undefined too?

gardnerjr:
interesting, it looks like the build times of old way and new way are just about the same?

was hoping maybe cloning just one repo for loc instead of 17 would save some time, but i guess it's cancelled out by the generation of loc.


i wonder if after this change the build could/should be split in 2:
1) one for the loc team to JUST to generate the resjson files, that they'd use as input to the loc pipeline -- this one is slow because of all the artifact uploading?
2) one for us/packaging that skips all that step and only consumes the localized things? this one would only update the output packages and would be quick?

would that end up with a 25 minute and a 10 minute build instead of 1 35 minute build?

gardnerjr (approved):
:shipit:

scripts/processDevPackageOnly.cmd
```diff
@@ -1,2 +0,0 @@
-powershell -NoProfile -ExecutionPolicy Unrestricted -Command "& '%~dp0processTemplates.ps1' dev"
-exit /B %ERRORLEVEL%
```
gardnerjr, line 2: what is the new version of this?

we use this in the testing instructions to allow people to generate a local gallery:
[https://github.com/microsoft/Application-Insights-Workbooks/blob/master/Documentation/Contributing.md#deploying-your-own-gallery](https://github.com/microsoft/Application-Insights-Workbooks/blob/master/Documentation/Contributing.md#deploying-your-own-gallery)

teams like WLI are using this to build an en-us only package they can deploy to storage to test/preview changes.  does the new process have a way to do this?
whatever it is you'll need to update the above doc/process as well.

ericajlin: Added back processDevPackageOnly.cmd that generates en-us package. Kept the name so I don't have to change the docs :)

VinothJeyaraman55 (approved):
Approving for Backup Report.workbook and usage-tab.workbook

austonli (approved):
AKS workbook changes LGTM

nibarnet (approved):

MichelleSwafford (approved):

blc16 (approved):

StefArroyo (commented):
#sign-off

PR #1287, opened 2021-02-25
Sign in logs adal

aiwangmicrosoft:
@danielwood95


aiwangmicrosoft:
@danielwood95 updated with some flavor text - do you think that makes sense or is there something else we should add?

danielwood95 (approved):

PR #1290, opened 2021-03-01
Update gitignore for new localization process

ericajlin:
Looks like LocProject got checked in accidently. This is automatically generated by the build so there's no need to check it in. Added file to gitignore.


yves-chan (approved):
LGTM

.gitignore
```diff
@@ -25,25 +25,8 @@ bld/
 [Ll]og/
 [Oo]utput/
 
-# localized repos that get cloned into scripts
```
gardnerjr, line 28: part of me says you might want to leave the old ones here for a little while?
otherwise, anyone whos run the scripts will suddenly have all these folders showing up as *additions * in git once they check out your new changes?

ericajlin: sure

PR #1291, opened 2021-03-01
[Kubelet Workbook] use new and old running pod count metric name in queries

gracewehner:
The metric `kubelet_running_pod_count` was renamed to `kubelet_running_pods` starting with kubernetes version 1.19. This breaks some queries for clusters using version 1.19. Changed queries to use both the < 1.19 and >= 1.19 name in queries to support both versions.


ganga1980 (approved):
lgtm

PR #1301, opened 2021-03-03
try to verify that templates don't have resource ids or subscription field

gardnerjr:
verify that templates don't have resource ids or even a subscriptions field with a guid value.

Adds another validation to `.armtemplate` files to look for a specific content like `"subscriptions": "[guid]"` and resource ids.

this isn't a perfect check as it is really specific to what endpoints are called by the template and how it is set up, but catches at least one case that was happening and required a hotfix.


test/jsonValidationTest.js
```diff
@@ -284,6 +286,17 @@ function validateNoResourceIds(settings, file) {
     }
 }
 
+function validateNoSubscriptionIdInTemplate(settings, file) {
+    // make sure there are no strings like "subscriptions": "[guid]" in the whole content. this is used on arm template files
+    // where it is more common to have a thing like this for certain endpoints.
+    // not parsing individual steps/etc at this time
+    let str = JSON.stringify(settings);
+    let regexp = /(\"subscription[s]*\"\s*:\s*\"(\/subscriptions\/)*[a-z0-9]{8}-[a-z0-9]{4}-[a-z0-9]{4}-[a-z0-9]{4}-[a-z0-9]{12})\"/gi;
```
yves-chan, line 294: This works for the one sub. What would the value be for multiple subscriptions? Would it be an array of subs or something?

gardnerjr: in this specific case, the value in the templates is parts of a resource id, so the literal string in the content was

```
    "subscriptions": "[the guid]",
    "resourceGroups": "[the rg name]",
    "resourceName": "[the name]",
    "resourceType":" "[the resource type]"
```

like how a resource id has `/subscriptions/` there.  you're right that this wouldn't match an array of subs, i was more trying to catch both the above case and anything like `subscription` or `subscriptions` as the field name and any string that was just a guid or resource id (though i guess the resource id check would also already find that case?)

gardnerjr: changed the regex to also look for arrays of subs as well.  it will fail on things like:
````
    "subscription": "00000000-0000-0000-0000-000000000000",
```

or 

````
    "subscriptions": [
           "00000000-0000-0000-0000-000000000000", "and moore"
    ]
```

if anyone has false positives we'll have to revisit

yves-chan (approved):

PR #1317, opened 2021-03-09
Modified the CODEOWNERS file to include the contact for Cosmos DB

StefArroyo:
## PR Checklist

Adding 

- Adding the cosmos team to the cosmos workbooks as CODEOWNERS.


gardnerjr (changes requested):
as is, this would make you (and you alone) the **only** codeowner.  you'd never be able to make any changes to this folder, since you can't approve your own things.

you should create a github team, that has you and at least one other member from cosmosdb, so you can approve your own things.  otherwise, you'll need to add another team, like ours, so that there's some other approvers.

only teams should be mentioned in this file, not individual users.  (any individual users are leftovers we're trying to clean up)

StefArroyo:
ok! That makes sense will update the PR.

gardnerjr (approved):

PR #1318, opened 2021-03-09
Code changes to handle error messages in NW provider

hsridharan:
## PR Checklist

Code changes to add error messages for the following scenarios :
1. When the Provider is not added, and the user tries to access the workbook
CreateProvider
2. When none of the NW providers are successfully created.  
ProviderNotcreated
3. When the Log Analytics tables are created but the data is not yet ingested.


PakDLiu:
For the newer version of SapMonitor, how did you determine if there is a new version?
Also, the error message should probably say "...deleting and creating a new SAP monitor"
And if this were to happen, will it still show the current workbook and metrics?

hsridharan:
@PakDLiu - I added SAPMON_VERSION column to the LA tables. Here in the WB, I am checking if the version is < version in LA table then I am showing the alert in the WB. I updated the error message as u recommended.

PakDLiu:
I am not sure if we should have a validation on the payload version.
We release new versions all the time (bug fixes, new providers) that may not have any affect on the netweaver provider.

hsridharan:
@PakDLiu - Yes, Ideally this message would never appear because most of the new AMS will have version 2.8 or more and quey will always pass. We'll add the queries to check the newly added column or anything related in the future when the need comes into the picture.

persiaAziz:
I don't think we can use the term "SAP Monitor" as the third warning above for legal reasons.

persiaAziz (changes requested):
Please make sure to not use term as "SAP Monitor" as it is not legally allowed

Workbooks/SapMonitor/SapNetWeaver/SapNetWeaver.workbook
```diff
@@ -8,6 +8,143 @@
...
+              "parameterName": "param_check_error",
+              "comparison": "isEqualTo",
+              "value": "2"
+            },
+            "name": "IngestionError"
+          },
+          {
+            "type": 1,
+            "content": {
+              "json": "We have a newer version of SAP monitor. If you would like to see more metrics, please upgrade by deleting and creating a new SAP monitor.",
```
persiaAziz, line 134: Why would they need to delete and deploy completely new monitor? We upgrade them automatically

persiaAziz (approved):

PakDLiu (approved):

PR #1319, opened 2021-03-09
Multicluster WB Update

austonli:
Minor fixes:
Changed dropdown style to Standard for consistency and updated template links to match the revised template names


austonli (commented):
I tested the changes, but the timeContext being 0 seems to not make a difference, since it's still using the time value from the timeRange parameter, not sure why it's listed as changed here.

yves-chan (approved):

PR #1328, opened 2021-03-11
Database and container selection

StefArroyo:
## PR Checklist

Added in the ability to select a database and container to filter out the view in the graphs. For this, added filters to each of the relevant charts to filter out by Container and Database depending on the selection of the user.

image

Also added a new chart that groups latency by operation type in the "Latency" tab.

image


Workbooks/CosmosDb/Resource Insights/Resource Insights.workbook
```diff
@@ -69,17 +69,65 @@
...
+      "conditionalVisibility": {
+        "parameterName": "Database",
+        "comparison": "isNotEqualTo"
+      },
+      "name": "parameters - 11"
+    },
+    {
+      "type": 1,
+      "content": {
+        "json": "{Database}"
```
yves-chan, line 127: Can take this text out, or can incorporate it into a workbook by formatting this that makes it more obvious that that they are looking at the scope of a particular DB/Container

Workbooks/CosmosDb/Resource Insights/Resource Insights.workbook
```diff
@@ -179,17 +227,31 @@
...
                 {
                   "namespace": "microsoft.documentdb/databaseaccounts",
                   "metric": "microsoft.documentdb/databaseaccounts-Requests-TotalRequests",
                   "aggregation": 7,
-                  "splitBy": "StatusCode"
+                  "splitBy": null
                 }
               ],
               "title": "Total Requests",
+              "filters": [
```
yves-chan, line 241: image
Can also play around with the idea of adding the DB/Container to the title? What do you think about that? If you want to format it to readable text on the title, you can also look at using criteria params and you can set rules lilke, if DB is empty, then the value is an empty string. If DB has a value, then the value is (Database: {Database})

Workbooks/CosmosDb/Resource Insights/Resource Insights.workbook
```diff
@@ -69,17 +69,65 @@
...
     },
+    {
+      "type": 9,
+      "content": {
+        "version": "KqlParameterItem/1.0",
+        "parameters": [
+          {
+            "id": "0abe0edd-1fb0-46fb-b7f2-6c0c21245fda",
+            "version": "KqlParameterItem/1.0",
+            "name": "Container",
```
yves-chan, line 104: Maybe we should rename this Param as CollectionName?

yves-chan (commented):
Can you double check that all the metrics items will apply the filters? I was looking through some of the metrics in Throughputs and Requests for example, and didn't see the filters being applied

Workbooks/CosmosDb/Resource Insights/Resource Insights.workbook
StefArroyo: In Cosmos DB terminology to users, we label collections as containers.

Workbooks/CosmosDb/Resource Insights/Resource Insights.workbook
```diff
@@ -69,17 +69,61 @@
...
+            "value": null,
+            "typeSettings": {
+              "additionalResourceOptions": [],
+              "showDefault": false
+            },
+            "queryType": 12
+          }
+        ],
+        "style": "pills",
+        "doNotRunWhenHidden": true,
```
yves-chan, line 118: I think we should remove this also, because the whole workbook stalls after we make the change to defaultValue

yves-chan, line 118: I think we should remove this also, because the whole workbook stalls after we make the change to defaultValue

Workbooks/CosmosDb/Resource Insights/Resource Insights.workbook
```diff
@@ -69,17 +69,61 @@
...
+            "name": "Database",
+            "type": 2,
+            "description": "When the value is \"unset\" or not selected, the data being displayed is at the account level.",
+            "query": "{\"version\":\"ARMEndpoint/1.0\",\"data\":null,\"headers\":[],\"method\":\"GET\",\"path\":\"{Resources}/sqlDatabases\",\"urlParams\":[{\"key\":\"api-version\",\"value\":\"2020-04-01\"}],\"batchDisabled\":false,\"transformers\":[{\"type\":\"jsonpath\",\"settings\":{\"tablePath\":\"$.value.*.properties.resource.id\",\"columns\":[]}}]}",
+            "value": null,
+            "typeSettings": {
+              "additionalResourceOptions": [],
+              "showDefault": false
+            },
+            "defaultValue": "value::1",
```
yves-chan, line 85: we should the remove the `defaultValue` part, because sometimes it 'chooses' a value in the background and it doesn't seem to make it appear on the dropdown (might be a bug on our end)

Workbooks/CosmosDb/Resource Insights/Resource Insights.workbook
```diff
@@ -1395,12 +1500,33 @@
...
+                  "valueParam": "Container"
+                }
+              ],
               "gridSettings": {
                 "rowLimit": 10000
               }
             },
             "name": "Total Requests by Operation Type"
+          },
+          {
```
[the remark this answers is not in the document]
StefArroyo: Hey @yves-chan, not sure which extra text you are referring to here.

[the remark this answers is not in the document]
yves-chan:
```
{
  "type": 1,
  "content": {
    "json": ""
  },
  "name": "text - 3"
}
          ```

yves-chan (approved):

aliuy (approved):

PR #1356, opened 2021-03-25
Update Server Availability.workbook

WKahnZA:
adding fix for duplicate RG names due to case being different eg. ResourceGroup and resourcegroup are the same but come up as two different RGs

## PR Checklist


Workbooks/Virtual Machines/Availability/Server Availability.workbook
```diff
@@ -274,5 +265,8 @@
       "name": "query - 0"
     }
   ],
+  "fallbackResourceIds": [
```
gardnerjr, line 268: you don't need this `fallbackResourceIds` section either, surprising it didn't warn in the validation build.  no harm here specifically but surprising nontheless.

Workbooks/Virtual Machines/Availability/Server Availability.workbook
```diff
@@ -182,22 +181,17 @@
               "columnMatch": "$gen_group",
               "formatter": 13,
               "formatOptions": {
-                "linkTarget": "Resource",
+                "linkTarget": null,
```
gardnerjr, line 184: did changing this fix something or was it unintentional?  do individual computer rows still show up as links?

Workbooks/Virtual Machines/Availability/Server Availability.workbook
WKahnZA: seems to be there every time I copy from the portal, i'll remove it

Workbooks/Virtual Machines/Availability/Server Availability.workbook
WKahnZA: unintentional, that did break the link good catch. fixed and resubmitting

gardnerjr (approved):
FYI: now that i look closer at the query changes, you could have just done `extend ResourceGroup=tolower(ResourceGroup) ` and not had to change any of the other column names/settings/etc

PR #1368, opened 2021-04-01
allow more fields when create work item and polish templates

MSKaiLian:
## PR Checklist

* Allow more fields when create a work items
* Polish templates
* Truncate string for exception messages


akkumari-ms (approved):

gardnerjr (approved):
i didn't look at the contents, only at the codeowners changes

PR #1385, opened 2021-04-08
Add correlation to user with long lookback of queries

radennis:
Like in Anita, they used to show users that are running queries with long lookback. can maybe correlate between the id in commands and queries.

Anita had the "Detailed View" Button.
It opens a grid (under the histogram) with columns for: User, application, Number of queries.
Once you clicked on a bar, the grid changed accordingly and showed the top users (by query count) who ran the queries for the selected lookback period (per day)

image

### DIFF
--- a/Workbooks/ADXCluster/AtResourceCache/ResourceCacheDetails.workbook
+++ b/Workbooks/ADXCluster/AtResourceCache/ResourceCacheDetails.workbook
@@ -465,6 +465,8 @@
               "size": 0,
               "showAnalytics": true,
               "noDataMessage": "There were no queries on this table",
+              "exportFieldName": "NumberOfDaysBefore",
+              "exportParameterName": "LookbackPeriod",
               "queryType": 0,
               "resourceType": "microsoft.kusto/clusters",
               "gridSettings": {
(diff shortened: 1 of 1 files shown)


radennis:
One question: the text of the query count is aligned right, how do I change it to be aligned left?

Workbooks/ADXCluster/AtResourceCache/ResourceCacheDetails.workbook
```diff
@@ -565,6 +567,130 @@
...
+              "parameterName": "DatabaseName",
+              "comparison": "isNotEqualTo"
+            },
+            "name": "Cache Header - Copy"
+          },
+          {
+            "type": 3,
+            "content": {
+              "version": "KqlItem/1.0",
+              "query": "let defaultRetention = 365d * 10;\r\nADXTableUsageStatistics \r\n    | where DatabaseName == '{DatabaseName}' \r\n    | where TableName == '{TableName}' \r\n    | extend NumberOfDaysBefore = datetime_diff('day', StartedOn, MinCreatedOn)\r\n    | where NumberOfDaysBefore == '{LookbackPeriod}'\r\n    | summarize Count=count() by User, ApplicationName\r\n    | top 10 by Count\r\n",
```
svilenko, line 589: I think we need to filter on timeRange as well. In other queries I added the timeRange from the parameter(It is only inserted today so not in production yet) WDYT?

guregini: @svilenko - yes.. It should be filtered the parameter
@radennis - for the alignment, you can use the "Text" column rendered property
image

liatbezalel:
Consider adding a text that explains the users they can click on the table row to get more details

Workbooks/ADXCluster/AtResourceCache/ResourceCacheDetails.workbook
guregini: and please change "LookbackPeriod" to "lookback period"
image

radennis: Why do we need to filter on timeRange if it's already filtered by the default retention?

svilenko: Maybe I am not looking correctly but I do not see filter by default retention in your query. Regardless, this query is running as a result of choosing a row in the table above. the table above is generated with the timeRange filter. If you do not filter this query as well we might get conflicting results which will be confusing for users.

svilenko: Also, we want all queries to be filtered somehow on timerange so the query would not be heavy

guregini (approved):

PR #1391, opened 2021-04-10
Updating api types

StefArroyo:
## PR Checklist

Made changes so that workbook can now detect the correct API type. This will allow the user to filter through a  Database and Collection depending on their database type. 

Also, changed the default time range to 4 hours and moved some tables to the bottom to match the at scale view.

image


Workbooks/CosmosDb/Resource Insights/Resource Insights.workbook
```diff
@@ -67,16 +67,93 @@
...
+                  "rightValType": "static",
+                  "rightVal": "Sql",
+                  "resultValType": "static",
+                  "resultVal": "sqlDatabases"
+                }
+              },
+              {
+                "criteriaContext": {
+                  "leftOperand": "EnabledApiTypes",
+                  "operator": "contains",
```
yves-chan, line 103: should the operators here just be ==?

yves-chan (approved):

Workbooks/CosmosDb/Resource Insights/Resource Insights.workbook
StefArroyo: The reason for the contains is that these API types (besides SQL) have both SQL and their respective API Type. For example, for Gremlin, you will see ("Sql","Gremlin").

aliuy (approved):

PR #1396, opened 2021-04-12
[CAE Workbook] Added new query widget and resolved minor parsing bugs

joelmas:
## PR Checklist
Added new widget/query to show CAE support status filterable bby App and Resource

Minor query updates to existing template
1. Incorrect parsing over AuthenticationProcessingDetails logs leading to wrong data being shown in the workbook. Regex fixed
2. Typo change in column name
3. As per new PM guidance, only showing CAE data if IpSeenByRp is different and not empty

* [X] post a screenshot of templates and/or gallery changes
* [X] ensure your template has a corresponding gallery entry in the gallery folder
* [X] ensure all steps have meaningful names
* [X] ensure all parameters and grid columns have display names set so they can be localized
* [X] ensure that parameters id values are unique __or they will fail PR validation__ (parameter ids are used for localization)
* [X] ensure that steps names are unique __or they will fail PR validation__ (step names are used for localization)
* [X] grep `/subscription/` and ensure that your parameters don't have any hardcoded resourceIds __or they will fail PR validation__
* [X] remove `fallbackResourceIds` and `fromTemplateId` fields from your template workbook __or they will fail PR validation__

Screenshots of Workbook:
image
image


joelmas:
image
image

abagaylelb (approved):

joelmas:
@microsoft/azure-ad-workbooks : Could anyone from the team please review/approve this workbook? It has been approved by the logsAnalytics team already.

joelmas:
@karanb22 : Would be able to approve this PR?

SarahBar (changes requested):
Took a look at your workbook. Requesting a few changes before approving:

1.) Change "IP Address mismatch seen by AAD and RP" section to Full size rather than Medium. Full size will remove the extra white space between that section and the next. As it is currently, the white space might prevent users from noticing there's a third section to scroll to.
2.) Default Time Range should be 24 hours (recommended because some customers' queries will time out if the time range is longer).
3.) When you have the Users parameter set as a drop-down, the drop-down menu will only contain a list of users who total up to 10,000 total sign-ins or less. Some tenants might contain more than 10k sign-ins, in which case some of the users would not be displayed in the drop-down. You do have the option to turn the drop-down into free-form text. There are tradeoffs to this choice, but if you want your users to have the option to filter to any user even for very large tenants with lots of sign-ins, then the free-form text box is the way to go. (Your choice, just raising this item for awareness.)
4.) In the "Continuous access evaluation Support Status" section, you have the Resource parameter set by default to Office 365 Exchange Microservices. Is this by design? If a user doesn't have this resource, then the section will be empty which might not be the best experience. Would it make more sense to select All by default? (Again, your choice but I'm raising it for awareness.)

Aside from these issues/considerations, it looks good.

joelmas:
@SarahBar : Thanks for the review. I appreciate all of the callouts. I definitely missed a few things that you did catch.
1. Change "IP Address mismatch seen by AAD and RP" section to Full
2. Updated Default range to 24 hours
3. Removed Users param since it would have created issues where we wouldn;t be able to spot IP mismatches. Thanks for explaining how the query works internally.
4. Set the default to All apps

SarahBar (approved):

PR #1429, opened 2021-05-03
Add sample for insights before users enable logs

svilenko:
Add sample for insights before users enable logs


liatbezalel:
what abount ingestion tab?

vutran01 (approved):

PR #1431, opened 2021-05-03
Moving MultiCluster folder to be under AKS

austonli:
- Moved all files/folders from MultiCluster to AKS/MultiCluster
- Updated the gallery schema to have AKS preceding the overview template
- Changed template links to controller and performance workbooks to include AKS in front in overview template


yves-chan:
Did you want to create copies of the old workbooks, since pins using the old workbooks won't have the same template reference?

gardnerjr:
closing ancient uncompleted pull requests

PR #1479, opened 2021-05-26
Support message updates

cartersocha:
Added error messages to help with support


Workbooks/Availability/Downtime Outages/Downtime Outages.workbook
```diff
@@ -272,7 +272,7 @@
           {
             "type": 1,
             "content": {
-              "json": "<h2>No Azure Subscriptions are selected.</h2>\r\n\r\nSelect one or more subscriptions to get started.\r\n\r\n**If you select more than 100 resources it will break the query**",
+              "json": "<h2>No available Azure Subscription has Webtests present or no subscription has been selected.</h2>\r\n\r\nEnsure that your selected subscriptions actually have Webtests present. You may have a subscription filter present. Select one or more qualifying subscriptions to get started.\r\n\r\n**If you select more than 100 resources it will break the query**",
```
SoubhagyaDash, line 275: Standardize to calling them "Availability tests" - all across this workbooks (and when updating other documentation). Web tests are too close to "VS multi-step web tests" for comfort + apparently exclude ping tests.

Users cannot end up in a state where "no subscription has been selected". Portal does not allow that. So this should just be "selected subscriptions do not contain application insights resources that have availability tests."

cartersocha: The portal does allow for no subscription to be selected as a workbook parameter. It allows for users to uncheck all parameters if they so choose. It could also indicate a query error of some sort. Updated the test naming.

gardnerjr: you can also get to that state if you have no subscriptions, or no subscriptions with any AI resources to pick from.

gardnerjr: There are also things in workbooks in multiselection to set a limit so that a user cannot select more than 100 subscriptions/resources, whatever. if you aren't doing that i'd suggest turning it on, and then maybe change the message not to "will break the query" but something more vague like "is not supported" ?

cartersocha: The 100 resource selection limit is already on. I updated to the not supported message.

MSKaiLian (approved):

PR #1496, opened 2021-06-08
Set required flag on hidden clusters list.

myunghojung:
## PR Checklist

HCI cluster picker should show loading icon when querying is in progress. But, it didn't when dependent list was empty. The change makes it required and adds dummy string to avoid error message in the other view using the parameter when the required parameter returns nothing.


yves-chan (approved):
Looks good!

muthus-msft (approved):

PR #1507, opened 2021-06-10
Adding new work for Azure AD cloud sync feature

dhanyahk:
## PR Checklist

Azure AD team is adding a new workbook to an existing gallery. The new workbook enables customers who use Azure AD cloud sync feature to sync identities from Active Directory to Azure Active Directory to get insights into those synced identities. IT Admins can easily get insights into how many identities got synced, how many failed and if they failed, why. They also get insight into details of identities that failed to sync as well as ones that are successful.

**screenshot of workbook**
image

**Gallery changes**

image


SarahBar (commented):
1. I notice that you have the default timeframe set to "Last 24 hours" which is great for query load time. But given the context of the report (especially the bar chart at the daily grain), a somewhat longer default timeframe might make more sense. I think you can bump it to "Last 3 days" without adverse query performance effects. 
2. The first data viz on the first tab is lacking context. Suggest adding a title and also thinking through how to make it clear what "user" and "group" means in the context of the counts and sparklines on the tiles. 
3. Consider adding some brief descriptive text at the beginning or on a separate tab, describing what the report shows and in what circumstances it is useful. 
4. Add a more informative label to the legend on "Sync Actions per day" (see Chart Settings > Series Settings). 
5. Check titles, column headings, etc. for consistent capitalization throughout. For example, "Sync Actions per day" should be "Sync Actions Per Day" if you are using title caps in headings/titles.
6. The "Unique Synced objects" column could be widened to show the whole column heading by default. At least, that's the case at my screen resolution. 
7. "Objects synced by Status" (on the 2nd tab, Cloud Sync Provisioning Log Details): when a status type is equal to zero, the tile for that status disappears. Is it possible to keep it visible with a displayed value of 0? (I did not find a way, but it's potentially misleading when there are status types not listed. Some descriptive text could potentially mitigate the issue here.)

dhanyahk:
1. Reason for 24 hours is the sync cycle is very frequent and having 3 days might make it too large a period to filter information (Especially the 2nd tab grid view which is based the time filter). Hence, the reason for keeping it as is.
2. Fixed
3.  Can I do this as part of the final release?
4. Done
5. Done
6. I can see it fully in mine
7. I will see what I can do here. Is this something I can address in final release?

SarahBar (approved):
Approved per discussion regarding preview for testing, followed by further changes for final release.

PR #1512, opened 2021-06-15
Sign in logs adal

aiwangmicrosoft:


aiwangmicrosoft:
@SarahBar Added non-interactive user sign-ins to the existing ADAL app list chart, also updated chart title to reflect that it includes both interactive and non-interactive user sign-ins only.

SarahBar (approved):
As discussed, approving. I will look into alternatives to the union in the KQL query for the ADAL chart, and thanks for looking into whether non-interactive sign-ins can/should be added to the other charts.

PR #1528, opened 2021-06-29
updates for RHEL cluster support

rsponholtz:
## PR Checklist

* [ *] Explain your changes, so people looking at the PR know *what* and *why*, the code changes are the *how*.
Contains query changes to take into account differences in data with RHEL clusters vs. SUSE clusters

* [ *] post a screenshot of templates and/or gallery changes
No differences apparent to users, simply adds RHEL cluster data format support.

* [ *] ensure all steps have meaningful names
* [* ] ensure all parameters and grid columns have display names set so they can be localized
* [* ] ensure that parameters id values are unique __or they will fail PR validation__ (parameter ids are used for localization)
* [ *] ensure that steps names are unique __or they will fail PR validation__ (step names are used for localization)
* [ *] grep `/subscription/` and ensure that your parameters don't have any hardcoded resourceIds __or they will fail PR validation__
* [ *] remove `fallbackResourceIds` and `fromTemplateId` fields from your template workbook __or they will fail PR validation__


Workbooks/SapMonitor/PrometheusHaCluster/PrometheusHaCluster.workbook
```diff
@@ -403,7 +407,8 @@
                         },
                         {
                           "columnMatch": "value_d",
-                          "formatter": 5
+                          "formatter": 5,
+                          "formatOptions": {}
```
shaaga, line 411: Are all formatOptions() supposed to be null ? If yes, why use them?

PakDLiu (approved):

Workbooks/SapMonitor/PrometheusHaCluster/PrometheusHaCluster.workbook
rsponholtz: The formatOptions were automatically added by the workbook editor.  As far as I know, there are no impacts to leaving them in - would you propose us editing the automatically-generated workbook json to remove these?  If so, I'd suggest having some type of processor that removes everything that should be removed when putting the json for a workbook in GitHub..

PR #1536, opened 2021-07-11
Added IoT Central workbooks

rangv:
## PR Checklist


Workbooks/IoT Central/IoT Edge device details/iotedge-device-details.workbook
```diff
@@ -0,0 +1,3347 @@
...
+      "content": {
+        "version": "KqlParameterItem/1.0",
+        "parameters": [
+          {
+            "id": "63aa6750-da91-4e11-af98-50e156843464",
+            "version": "KqlParameterItem/1.0",
+            "name": "SelectedDevice",
+            "label": "💻 Choose device",
+            "type": 2,
+            "query": "\nlet startTime = {timeRange:start};\nlet endTime = {timeRange:end};\nlet devices = () { \nInsightsMetrics\n| where Name == \"edgehub_gettwin_total\"\n| where TimeGenerated > startTime and TimeGenerated < endTime\n| extend dimensions=parse_json(Tags)\n| extend device = tostring(dimensions.edge_device)\n| distinct device\n};\ndevices()",
```
gardnerjr, line 178: FYI, for the time part, you don't need to use let to get start/end.  there's also a default syntax that will be generated if you do

```
where TimeGenerated {timeRange}
```

which will generate 

```
where TimeGenerated between( "something" .. "something else" )
``

unless you need start/end, this syntax can generally make for much simpler queries

The real reason i commented here though is that this query fails on an iot app that isn't onboarded:

![image](https://user-images.githubusercontent.com/10158007/125985257-6308d732-cada-4507-ab61-131708e0fa77.png)


you're already hiding/showing other things if not onboarded, not sure if this was intentional or missed here?

File: Workbooks/IoT Central/IoT Edge health snapshot/iotedge-health-snapshot.workbook [hunk 2]
```diff
@@ -0,0 +1,872 @@
...
+      "conditionalVisibility": {
+        "parameterName": "neverset",
+        "comparison": "isNotEqualTo"
+      },
+      "name": "parameters - 7"
+    },
+    {
+      "type": 1,
+      "content": {
+        "json": "## This workbook is designed to be invoked from the IoT Edge Fleet view workbook.",
```
gardnerjr, line 69: if this is true, you don't actually need to expose it at all, you can leave it here, and remove this entry from the gallery.  things that are used "inside" other things don't need to be exposed at the top level in the gallery

gardnerjr (approved):
i only have one iot app, and it isn't onboarded, so i can't verify the full behavior of the workbooks, but i'm also not an expert in the area so i wouldn't know what is what anyway.

i'm only really verifying that they are somewhat functional and really to check the CODEOWNERS changes.  someone from your team should also be reviewing for content/etc if they haven't already.

PR #1545, opened 2021-07-19
add docs, sample for using toolbar style

gardnerjr:
Adding docs for the new toolbar style in the links step.

Also includes a sample that uses globals, toolbar, and arm actions to start/stop websites.


renu00 (approved):
Looks good!

PR #1549, opened 2021-07-21
Change aggregation type to max for used memory

brendanzagaeski:
Because the Azure Cache for Redis used memory metrics track single
values over time, aggregating by sum resulted in graphs that showed
higher than expected values for time granularities larger than 1 minute.
For example, if the used memory for a cache was constant at 100 MB, the
graph showed a value of 200 MB when set to a time granularity of 1
minute and 1 GB when set to a time granularity of 5 minutes.

To solve this issue, change the aggregation type from sum to max. This
results in graphs that show the expected values for all time
granularities. It also aligns the workbook with the at-scale workbook
and the Overview section of Azure Cache for Redis resources in the Azure
portal, which both already aggregate used memory by max.

At the moment, some places in the Azure portal still link to the
workbook via the old Workbooks/Resource Insights path, so apply the
change under both the old path and the new path.

## PR Checklist

  | Before | After |
  | :------: | :-----: |
  | image | image |
  * N/A, no changes
  * N/A, no changes
  * N/A, no changes
  * N/A, no changes
  * N/A, no changes
  * N/A, no changes
  * N/A, no changes
  * N/A, no changes

### DIFF
--- a/Workbooks/RedisCache/AtResource/ResourceInsights.workbook
+++ b/Workbooks/RedisCache/AtResource/ResourceInsights.workbook
@@ -353,53 +353,53 @@
                 {
                   "namespace": "microsoft.cache/redis",
                   "metric": "microsoft.cache/redis--usedmemory0",
-                  "aggregation": 1,
+                  "aggregation": 3,
                   "splitBy": null
                 },
                 {
                   "namespace": "microsoft.cache/redis",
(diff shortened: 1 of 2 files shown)


alfantp (approved):
:shipit:

brendanzagaeski:
**Changes in latest force-push:**

Rebase on master.

PR #1553, opened 2021-07-23
Change aggregation types for RedisCache metrics

brendanzagaeski:
* Add display names for the server load, CPU, used memory, and connected
  clients tiles to indicate that they show max values.
* Change the tiles for errors, expired keys, and evicted keys to
  aggregate by sum to match the larger graphs for those metrics.
* Change the connected clients graph to aggregate by max instead of sum.
  Each connected clients data point records the total number of
  connected clients, not the change in number of connected clients, so
  aggregating by max or by average is typically the most useful option.
  Aggregating by max helps reveal temporary spikes of high activity.
* Change the graphs for cache read, cache write, and operations per
  second to aggregate by max instead of sum. Each of these metrics
  records values in units per second. Although it's tempting to
  aggregate by sum to try to get total numbers over larger time
  intervals, that doesn't quite work because the metrics are updated
  more often than once per second. Instead, aggregate by max to help
  reveal temporary spikes, like for connected clients.
* Change the tiles for the cache read, cache write, operations per
  second metrics to aggregate by max as well, and update the
  corresponding display names.
* Standardize all of the graphs to use the `TimeRange` parameter for the
  time context so that setting the parameter will update all of the
  graphs.

Like in commit f7bde430, apply the changes under both the old and new
path for the workbook.

## PR Checklist

  **Overview**
  | Before | After |
  | :------: | :-----: |
  | overview-before | overview-after |

  **Performance**
  | Before | After |
  | :------: | :-----: |
  | performance-before | performance-after |

  **Operations**
  | Before | After |
  | :------: | :-----: |
  | operations-before | operations-after |

  * N/A, no changes
  * N/A, no changes
  * N/A, no changes
  * This PR makes a few small adjustments to add the aggregation type "(Max)" in the display name of some of the tile grid columns.
  * N/A, no changes
  * N/A, no changes
  * N/A, no changes
  * N/A, no changes

### DIFF
--- a/Workbooks/RedisCache/AtResource/ResourceInsights.workbook
+++ b/Workbooks/RedisCache/AtResource/ResourceInsights.workbook
@@ -150,37 +150,41 @@
                 {
                   "namespace": "microsoft.cache/redis",
                   "metric": "microsoft.cache/redis--serverLoad",
-                  "aggregation": 3
+                  "aggregation": 3,
+                  "columnName": "Server Load (Max)"
                 },
                 {
                   "namespace": "microsoft.cache/redis",
                   "metric": "microsoft.cache/redis--percentProcessorTime",
-                  "aggregation": 3
+                  "aggregation": 3,
+                  "columnName": "CPU (Max)"
                 },
                 {
                   "namespace": "microsoft.cache/redis",
                   "metric": "microsoft.cache/redis--usedmemory",
-                  "aggregation": 3
+                  "aggregation": 3,
+                  "columnName": "Used Memory (Max)"
                 },
                 {
                   "namespace": "microsoft.cache/redis",
                   "metric": "microsoft.cache/redis--connectedclients",
-                  "aggregation": 3
+                  "aggregation": 3,
+                  "columnName": "Connected Clients (Max)"
                 },
                 {
                   "namespace": "microsoft.cache/redis",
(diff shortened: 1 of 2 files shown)


Workbooks/RedisCache/AtResource/ResourceInsights.workbook
```diff
@@ -582,7 +589,7 @@
               ],
               "timeContextFromParameter": "TimeRange",
               "timeContext": {
-                "durationMs": 14400000
```
alfantp, line 585: Intentional? What does it mean to change this?

brendanzagaeski: Good question. I included the change for this `durationMs` to make the value consistent with the other graphs that included `"timeContextFromParameter": "TimeRange"`. In particular, I noticed that in the graphical editor for the workbook in the portal, when I changed `timeContextFromParameter` to `TimeRange`, it also saved a `0` for `durationMs`. My understanding is that when the time context is configured to read from a parameter, it overrides the `durationMs` property, so that property can just be `0` for simplicity.

alfantp (approved):
:shipit:

brendanzagaeski:
**Changes in latest force-push:**

Rebase on master.

PR #1559, opened 2021-07-29
common parameters and availability tab

kavitha2270:
## PR Checklist

1. Removed old code
2. Created a common parameter that has a function, which takes the table name and verifies if the table exists, If so column exists, If so records exists and returns the result. This function is called number of times as the table count on the log analytics. The derived results from these function calls are concatenated as part of single string with table name and status. This string is later used in workbook for verification purpose for the tab metrics visibility. This change attributed in removing code duplication and making it more performant.
3. Moved "param_check_provider" parameter to the middle of the page that fixed the info message popping up before the page loads.
4. Merged with Devansh's change for common parameters across all tabs.(USER STORY 7616429).
5. Introduced Availability tab after feedback from customers and team members.
6. Timestamp changes were handled differently for SOAP and RFC metrics calls(SOAP -serverTimestamp_t and RFC-timestamp_t).

Testing

1. Tested for RFC metrics and non RFC metrics with common parameter selection across all tabs.
2. Tested in dev and production subscriptions
Capture
Capture2


hsridharan (approved):

hsridharan (approved):

PakDLiu:
Can you post screenshots of the new changes?

PakDLiu (approved):

PR #1564, opened 2021-08-05
New look for Azure AD cloud sync workbook

dhanyahk:
Incorporated Design changes to Workbook

## PR Checklist

* Incorporated new look (design changes) to Azure AD cloud sync workbook. 
* Incorporated best practices with Design team's help to provide a consistent and professional look to the workbook
AD2AAD_Workbook_Screenshots
AD2AAD_Workbook_Screenshots_Details

### DIFF
--- a/Workbooks/Azure Active Directory Provisioning/AAD cloud sync Insights (Preview)/AAD cloud sync Insights.workbook
+++ b/Workbooks/Azure Active Directory Provisioning/AAD cloud sync Insights (Preview)/AAD cloud sync Insights.workbook
@@ -4,15 +4,22 @@
     {
       "type": 1,
       "content": {
-        "json": "# Azure AD cloud sync insights (AD2AAD)\r\n"
+        "json": "# Azure AD cloud sync insights (AD2AAD)"
       },
       "name": "text - 2"
+    },
+    {
+      "type": 1,
+      "content": {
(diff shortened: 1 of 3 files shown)


SarahBar (approved):
Nice changes coming out of the collaboration with UX. Approving with some minor issues to look into, and you could possibly submit a new PR (or save these changes to be applied to the next revision):
- Change "light weight" to "lightweight" (one word)
- Status drop-down menu allows multi-select, but if you select more than one option, the query does not work
- Is 48 hours a sensible default time range? In Woodgrove, I need to go back 60 days to start seeing query results. What is the minimum time range that would be sufficient to start seeing data in a non-test environment?

dhanyahk:
> Nice changes coming out of the collaboration with UX. Approving with some minor issues to look into, and you could possibly submit a new PR (or save these changes to be applied to the next revision):
> 
> * Change "light weight" to "lightweight" (one word)
> * Status drop-down menu allows multi-select, but if you select more than one option, the query does not work
> * Is 48 hours a sensible default time range? In Woodgrove, I need to go back 60 days to start seeing query results. What is the minimum time range that would be sufficient to start seeing data in a non-test environment?

48 hours usually is good since sync happens every day at some level. If I change the time to a greater range, it may not be useful. Maybe I can do 7 days. 

will fix the status drop down.

PR #1600, opened 2021-09-01
Remove preview tag from Risky IP Workbook

zhvolosh:
Publish for customers by removing preview tag + change name of Sign-In Analysis to display AAD and ADFS information

## PR Checklist

* [X] Explain your changes, so people looking at the PR know *what* and *why*, the code changes are the *how*.

Removing the preview tag from the Risky IP Report Workbook to have it be available to customers + changing name of Sign-Ins Workbook to show AAD+ADFS title. 


zhvolosh:
@microsoft/azure-ad-workbooks can someone take a look and merge this request? I'm looking to release this Workbook to customers without the preview tag after validating in the past Q.

SarahBar (changes requested):
Before merging the Risky IP Workbook PR, could you do a few things?

(1) Add the explanation at the beginning to call out the need for ADFSSignInLogs stream and how to make it available for Workbooks to consume.

(2) In the "Risky IP Count by time" chart, make the y axis count only whole numbers from 1 (1, 2, 3...). Currently, it shows decimal increments, which doesn't make sense because you can't have a fraction of an IP address counted.

(3) Optional but recommended: Add more filtering options to the tables--for example, IP Address, Location, User Name, User ID, Error Code and/or Error Description. You can choose the ones that you think make the most sense given what users are likely to be searching for. Especially in the case that a table can only return a limited number of results, the filters make it possible to find something that would otherwise be hidden.

zhvolosh:
Addressed the following:
1. Added a toggle tab to "Show Help" with prerequisite details to enable Log Analytics.

2. I think this one cannot be changed to not have decimals as the 3 options are decimal, percent, and currency. I played around with the settings and there is a limit where I cannot set a value of zero for the minimum significant digits. If the range is bigger, and not all the values are 1, it should shift to be correct since it is automatic.
image

3. I added User Name to the Select By IP Table, which would also make sense in the filtration.

SarahBar (approved):

PR #1654, opened 2021-10-13
Add max aggregation for CPU in VM details

BernieWhite:
## PR Checklist

Currently the VM details workbook includes a metric for Percent CPU using the _avg_ aggregation. When checking the performance of VMs it is often helpful to understand what the _max_ of a datapoint is. For example if the granularity is 8 hours, an average could quite different then as represented by a 5 minute granularity.

This PR:

- Updates the existing workbook to add a max metric aggregation.
- Removes the hardcoded theme to default to auto since there is two metrics displayed instead of one.


BernieWhite:
@nasundar Bump.

BernieWhite:
@nasundar @microsoft/applicationinsights-devtools Is there anything else that I need to do here to get this merged?

yves-chan:
@BernieWhite, Can you include some screenshots for before and after? We need to check with the authors of this workbook to ensure this change makes sense on a template level.

@nasundar @rakshith210 Can you review?

nasundar:
@BernieWhite sorry, I missed this PR and thanks @yves-chan for reminding me.
BTW, I am not the author of this workbook, looks like its @acearun. I have added Arun as reviewer as well.
The change looks pretty straight forward, and it makes sense to view "Max" for CPU. So, I will approve this PR. 

Thinking about this workbook more holistically, it might be better to provide a dropdown with supported aggregates so that user can pick the aggregates that suits their scenario, something similar to below. But this requires a separate discussion. 
image

nasundar (approved):

raghavthind2005 (approved):

PR #1713, opened 2021-12-10
Add Delivery Optimization template workbook for Update Compliance in Windows

cmknox:
Add a new Update Compliance template workbook for Delivery Optimization in Windows. 

Customers can use this template to monitor how Delivery Optimization is working in their environments.

DO_template_workbook_screenshot


Workbooks/WaasUpdateInsights/DeliveryOptimization/DeliveryOptimization.workbook
```diff
@@ -0,0 +1,366 @@
...
+            "version": "KqlParameterItem/1.0",
+            "name": "_SnapshotTime",
+            "label": "Snapshot Time",
+            "type": 1,
+            "description": "Time indicating the latest snapshot",
+            "isRequired": true,
+            "query": "WUDOAggregatedStatus | project TimeGenerated\r\n| union WaaSUpdateStatus | project  TimeGenerated \r\n| where TimeGenerated >= ago(28d)\r\n| summarize max(TimeGenerated)",
+            "isHiddenWhenLocked": true,
+            "timeContext": {
+              "durationMs": 86400000
```
gardnerjr, line 20: the query above has >=ago 28d but right here it looks like the time range for this query is fixed to last 24h?  if this is truly supposed to be 28d then you'd have to set this query's time range to "set in query" in the parameter ux.  Also, if that is true, then this could also be a very expensive initial query and this query has to finish before anything else can start since all the other queries depend on it? (do these tables possibly have lots of rows?  most workspaces have their hot cache set to 14d so anything outside that could be really expensive?)

i don't know how any of your things specifically work, this is just a general comment when i see large time ranges by default in queries.  customers with huge amounts of data in tables could have a really bad time?

gardnerjr: and conversely, if there's NO data in the tables (one of the workspaces i tried it on, then you get this:

image


the query returns no rows, so this required param has no value, and the rest of the workbook looks like this:
image


where it tells users to set a parameter they can't see, and can't fix?

Workbooks/WaasUpdateInsights/DeliveryOptimization/DeliveryOptimization.workbook
```diff
@@ -0,0 +1,366 @@
...
+      "name": "text - 2"
+    },
+    {
+      "type": 3,
+      "content": {
+        "version": "KqlItem/1.0",
+        "query": "let _SnapshotTime = datetime({_SnapshotTime});\r\n\r\nWUDOAggregatedStatus| where TimeGenerated >= _SnapshotTime\r\n| where ContentType in (\"Feature and Flight Updates\" , \"Quality Updates\"  ) | summarize BytesFromPeers = round((sum(0.0 + BytesFromPeers + BytesFromGroupPeers)/ sum(BytesFromPeers + BytesFromGroupPeers+BytesFromCDN)) * 100.0 ,2)",
+        "size": 4,
+        "title": "% bandwidth savings in Feature and Quality Updates over the last 28 days",
+        "timeContext": {
```
gardnerjr, line 63: there seem to be at  least a couple other places that seem to have time range set to 24h instead of "Set in query" that you'll want to verify?

Workbooks/WaasUpdateInsights/DeliveryOptimization/DeliveryOptimization.workbook
cmknox: We updated the query to look for 7 days and the time windows is now 7 days as well. 

We don't need the customer to set the Snapshot Time. Ideally we should have data within 7 days, if not, it is beyond our SLA and we'd be working on fixing such an issue. This is consistent with other workbooks in the WaasInsights solution.

Workbooks/WaasUpdateInsights/DeliveryOptimization/DeliveryOptimization.workbook
cmknox: Fixed, we are setting the time in the query itself.

natasha-tahariya (approved):

gardnerjr (approved):
in the next pr it looks like maybe the gallery file needs to have a codeowners entry so that our team is no longer required?

PR #1726, opened 2021-12-27
WorkbooksCreation

osalzberg:
I have created a workbook to replace the old solution of activity logs. as the owner of Activity logs from LA side.

here is a ppt of the project
https://microsoft.sharepoint.com/:p:/t/azureteams/docs/EWBYFSCOh_tFqrw3lCsf2bUBTWyzudmpxF54mSa1aabcxw?e=96dq8W


gallery/ActivityLogsInsights/Azure Monitor.json
```diff
@@ -0,0 +1,18 @@
...
+	"categories": [
+		{
+			"id": "ActivityLogsInsights",
+			"name": "Activity Logs Insights",
+			"templates": [
+				{
+					"id": "Workbooks/ActivityLogsInsights/AtScale",
+					"author": "Microsoft",
+					"description": "Activity Logs Insights",
+					"name": "Activity Logs Insights"
```
gardnerjr, line 13: do you want this one to be marked isPreview here too?

gardnerjr (approved):
since you have it marked preview, i haven't looked in detail at the workbook content, but am approving so you can get this in and start "bashing" it effectively as a preview.  (there's one place you didn't mark it preview, not sure if that was intentional or not

you might want to look at other kinds of checks or putting your "note x is required" not right at the top though.  that's wasting valuable real-estate for the people who have this hooked up.  you might want to do other kind of param or other related checks and use conditional visibility to show extra content like that?

other nit: THERES ALL CAPS TITLES IN PLACES and the portal generally moved away from that style?

PR #1790, opened 2022-02-16
Validation: Update Gallery File Validation Test

ericajlin:
Allow numeric characters in gallery files as some resource types contain a version number (eg. microsoft.dbforpostgresql-servergroupsv2


yves-chan (approved):
Also includes capitals A-Z, does that matter here?

gardnerjr (commented):
`\w` also includes `_` which i think we didn't want?

might be better to just add `0-9` to this instead of switching to `\w`?

gardnerjr (approved):

PR #1809, opened 2022-02-23
Onboard ACS team workbook

hirenchalodiya1:


gardnerjr (commented):
Is there any way to simplify the folder hierarchy?  it looks like the files are ending up with really long names and really deep paths?

`Workbooks/Azure Center for SAP solutions/SAP Inventory Checks/SAP Inventory Checks.workbook` is 92 characters long, with lots of spaces.  

None of these are *technical* limitations but might just be simpler if it were `SAP Solutions` or something shorter here in the repo?  (it would also sort it near the other SAP things?)

gardnerjr (approved):
I don't appear to have access to any resources to make the report work, and also don't know if the show help/report things were supposed to have any affect, they didn't for me?

Its also strange that there's a parameter to pick subscriptions, but it is hidden and defaults to all?  Shouldn't this expose that so a user could filter down?

Someone from your team should also be looking at the content and making sure it works, both in the "happy path" of having applicable resources, AND in the alternate paths for users like me who don't have any resources/etc and the report is essentially empty, to make sure it works for all users.

hirenchalodiya1:
This is seed file, we will update this workbook as we go

PR #1816, opened 2022-02-27
new workbooks for defender-for-iot

garubin:
## PR Checklist

We're adding new templates for our service, to have an overview about Alerts & Devices.
* [X] Explain your changes, so people looking at the PR know *what* and *why*, the code changes are the *how*.

* [X] post a screenshot of templates and/or gallery changes
* [X] ensure your template has a corresponding gallery entry in the gallery folder
* [X] If you are adding a new template, add your team and template/gallery file(s) to the CODEOWNERS file. CODEOWNERS entries should be teams, not individuals
* [X] ensure all steps have meaningful names
* [X] ensure all parameters and grid columns have display names set so they can be localized
* [X] ensure that parameters id values are unique __or they will fail PR validation__ (parameter ids are used for localization)
* [X] ensure that steps names are unique __or they will fail PR validation__ (step names are used for localization)
* [X] grep `/subscription/` and ensure that your parameters don't have any hardcoded resourceIds __or they will fail PR validation__
* [X] remove `fallbackResourceIds` and `fromTemplateId` fields from your template workbook __or they will fail PR validation__

Alerts workbook:
image
image

Devices workbook:
image

Health workbook:
image


gardnerjr:
general comments:

1) avoid using 30 days as the default time range - Log Analytics by default has a hot cache of 14 days or less, queries over wider ranges than that will require bringing cold data into the cache.  set your default time range to the smallest/fastest usable thing.  Most places use 24h as the default if they can to limit the initial query to a small set of data

2) or your long tail bar charts, you might consider switching to grid view and using the "bar" renderer on that column, and turn on filtering?  that would let you see/filter interactively instead of having the items try to squish into charts?

garubin:
> general comments:
> 
> 1. avoid using 30 days as the default time range - Log Analytics by default has a hot cache of 14 days or less, queries over wider ranges than that will require bringing cold data into the cache.  set your default time range to the smallest/fastest usable thing.  Most places use 24h as the default if they can to limit the initial query to a small set of data
> 2. or your long tail bar charts, you might consider switching to grid view and using the "bar" renderer on that column, and turn on filtering?  that would let you see/filter interactively instead of having the items try to squish into charts?

1. Changed.
2. The “long tail” bar charts happen because we are in a testing environment with dozens of sensor versions. In a production env it would look much better

yairzarka (approved):

PR #1819, opened 2022-02-28
Users/tolee/add tsg templates to workbooks

tonykslee:
copy the 4 templates that only show up in the TSG area to the Workbooks area

image


gardnerjr (changes requested):
nit: i'd put this tsg section at the bottom, not up here =

gardnerjr (approved):

PR #1861, opened 2022-04-11
CNS IP Metrics Workbook.

rsagasthya:
## Changes:
1. CNS emits certain metrics relating to the IP usage and three of these metrics, namely: 
 * cx_ipam_total_ips
 * cx_ipam_available_ips
 * cx_ipam_pod_allocated_ips
2. Creating a workbook to show these metrics according to:
 * the time period (required) specified by the user, and 
 * segregated by Node, if specified by the user. Default is to display the data of all nodes as a stacked bar graph.

## PR Checklist


vishiy (commented):
Approving after going thru the charts with Rahul. I see the default xa-xis legends overlap when a single node is selected in node filter. Please fix that.

image

vishiy (approved):

PR #2007, opened 2022-09-16
Adding Preview Workbook Templates for Azure Percept

nidhunge:
## PR Checklist

Added three preview workbook templates for Azure Percept. See screenshots attached.

image
image
image
image


gardnerjr (approved):
i don't have access to any data/resources to validate on, so i've really only scanned the workbook content.

PR #2096, opened 2023-01-07
Added a new tab to the workbook to list ip addresses in use

Khushbu-Parekh:
Added a new tab to list all the ips in use 

image


Workbooks/AKS/Subnet IP Usage/Subnet IP Usage.workbook
```diff
@@ -418,7 +426,7 @@
                   "type": 2,
                   "isRequired": true,
                   "query": "InsightsMetrics\n| where Name contains \"cx_ipam_total_ips\"\n| where Namespace == \"prometheus\"\n| extend jsonprop = parse_json(Tags)\n| extend hostname = tostring(jsonprop.hostName)\n| distinct hostname\n| project hostname",
-                  "value": "aks-nodepool1-25254241-vmss000003",
+                  "value": "aks-nodepool1-36771451-vmss000001",
```
gardnerjr, line 429: fyi you might want to have this query return a selected column to auto-select something?  right now, it is trying to restore this specific named value for everyone using the template, but will really only work for you?

vishiy (approved):

gardnerjr (approved):
the things i noticed looks like they got fixed

PR #2105, opened 2023-01-24
Waf workbook

sowmyam2019:
## PR Checklist

Adding a Azure Web Application Firewall workbook. Includes:
1) Creating directory in workbook called Azure Web Application Firewall
2) Creating directory in gallery called azure-web-application-firewall
3) added WAF Monitoring.workbook to workbook/Azure Web Application Firewall directory
4) added two json files in gallery folder since WAF has two RPs called microsoft.network-applicationgatewaywebapplicationfirewallpolicies.json and microsoft.network-frontdoorwebapplicationfirewallpolicies.json
5) updated codeowners file with alias of team for workbooks and gallery files.

* [ x] ensure your template has a corresponding gallery entry in the gallery folder
* [x ] If you are adding a new template, add your team and template/gallery file(s) to the CODEOWNERS file. CODEOWNERS entries should be teams, not individuals
* [x ] ensure all steps have meaningful names
* [x ] ensure all parameters and grid columns have display names set so they can be localized
* [ x] ensure that parameters id values are unique __or they will fail PR validation__ (parameter ids are used for localization)
* [x ] ensure that steps names are unique __or they will fail PR validation__ (step names are used for localization)
* [x ] grep `/subscription/` and ensure that your parameters don't have any hardcoded resourceIds __or they will fail PR validation__
* [x ] remove `fallbackResourceIds` and `fromTemplateId` fields from your template workbook __or they will fail PR validation__


Workbooks/Azure Web Application Firewall/WAF monitoring/WAF Monitoring.workbook
```diff
@@ -44,7 +42,7 @@
               "value::all"
             ],
             "value": [
-              "/subscriptions/27cafca8-b9a4-4264-b399-45d0c9cca1ab"
+              ""
```
gardnerjr, line 45: instead of empty string, should remove the value field+value entirely in a bunch of these place

Workbooks/Azure Web Application Firewall/WAF monitoring/WAF Monitoring.workbook
```diff
@@ -105,8 +103,8 @@
             "type": 4,
             "isRequired": true,
             "value": {
-              "durationMs": 2764800000,
-              "endTime": "2022-12-30T21:32:00.000Z"
+              "durationMs": 5707080000,
+              "endTime": "2023-01-06T18:50:00.000Z"
             },
```
gardnerjr, line 108: this appears to be a custom time range?  like from today back a certain amount of time, a previous version was looking for specific times in december?  i presume you really want this to be something like "last 24 hours", not a specific start and end time?

Workbooks/Azure Web Application Firewall/WAF monitoring/WAF Monitoring.workbook
```diff
@@ -192,23 +190,43 @@
...
...POLICIES\") and (\"{WAFType:label}\" == \"All\" or (ResourceType == \"APPLICATIONGATEWAYS\" and \"{WAFType:label}\" contains \"application gateway\") or (ResourceType == \"FRONTDOORS\" and  \"{WAFType:label}\" contains \"azure front door\") or (ResourceType == \"PROFILES\" and \"{WAFType:label}\" contains \"azure front door premium\") or (ResourceType==\"CDNWEBAPPLICATIONFIREWALLPOLICIES\" and \"{WAFType:label}\" contains \"cdn\")) and (\"{WAF:label}\" == \"All\" or Resource in ({WAF}))and (policy_s in ({WAFPolicy}))\r\n| where Category == \"FrontdoorWebApplicationFirewallLog\" or Category == \"FrontDoorWebApplicationFirewallLog\" or OperationName == \"ApplicationGatewayFirewall\" or Category == \"WebApplicationFirewallLogs\"\r\n| extend Rule = strcat(ruleName_s, Message), ClientIP = strcat(clientIp_s, clientIP_s)\r\n// Application Gateway has Matched, Blocked, Detected : translates to Matched, Block, Log\r\n// Azure Front Door has Matched, Block, Log : translates to Matched, Block, Log\r\n| extend Action = iif(action_s == \"Blocked\", Action = \"Block\", action_s)\r\n| extend Action = iif(Action == \"Detected\", Action = \"Log\", Action)\r\n| summarize number = count() by Action",
```
gardnerjr, line 226: these are filtering on a parameter *label* value, not an underlying value?  (in some context, things like labels are translated, and filtering on a label value might not be what you want?)

Workbooks/Azure Web Application Firewall/WAF monitoring/WAF Monitoring.workbook
```diff
@@ -0,0 +1,833 @@
...
+        },
+        "metrics": [
+          {
+            "namespace": "microsoft.cdn/profiles",
+            "metric": "microsoft.cdn/profiles--WebApplicationFirewallRequestCount",
+            "aggregation": 1,
+            "splitBy": null
+          }
+        ],
+        "title": "Sandy Corp Applications Request Count",
```
gardnerjr, line 743: is this title "sandy corp applications" expected? that doesn't seem like a generic title?

Workbooks/Azure Web Application Firewall/WAF monitoring/WAF Monitoring.workbook
```diff
@@ -0,0 +1,831 @@
+{
+  "version": "Notebook/1.0",
+  "items": [
+    {
+      "type": 1,
+      "content": {
+        "json": "## Sandy Corp Organization Events",
```
gardnerjr, line 7: this is even right at the top, all customers would have this same title, is this really what you want?

Workbooks/Azure Web Application Firewall/WAF monitoring/WAF Monitoring.workbook
```diff
@@ -0,0 +1,831 @@
...
+              "durationMs": 0
+            },
+            "timeContextFromParameter": "TimeRange",
+            "queryType": 0,
+            "resourceType": "microsoft.operationalinsights/workspaces"
+          },
+          {
+            "id": "1c584e06-9f2c-4d32-ab9f-bfef9b5ef425",
+            "version": "KqlParameterItem/1.0",
+            "name": "WAFPolicy",
```
gardnerjr, line 197: this parameter doesn't have a workspace set, and isn't set to use the Workspaces parameter like other params are?

is this intentional?

Workbooks/Azure Web Application Firewall/WAF monitoring/WAF Monitoring.workbook
```diff
@@ -0,0 +1,831 @@
...
+                }
+              ],
+              "allowCustom": true
+            }
+          },
+          {
+            "id": "604a42a0-deca-4a95-a15f-8977646a7fac",
+            "version": "KqlParameterItem/1.0",
+            "name": "WAFType",
+            "type": 5,
```
gardnerjr, line 140: this parameter is set to be a "resource picker" but does not return azure resources, it returns strings.  it even will show a warning like this inthe editor:
image
i think you want the parameter type here to just be "dropdown"

Workbooks/Azure Web Application Firewall/WAF monitoring/WAF Monitoring.workbook
```diff
@@ -0,0 +1,831 @@
...
+            "crossComponentResources": [
+              "{Workspaces}"
+            ],
+            "value": [
+              "value::all"
+            ],
+            "typeSettings": {
+              "additionalResourceOptions": [
+                "value::all"
```
gardnerjr, line 154: if you want "All" to be the default value, you can also set that in the settings, `Default Selected item: then pick "all"`
image

gardnerjr: that screenshot right there also lets you assign a specific value for "All", so you could use "*" or something instead of testing the *label value* for "All" which will not be "All" in all languages.

gardnerjr: (effectively all of these queries would use something like

`... "{WAFType}" == "*" ...` instead of `"{WAFType:label}" == "All"`)

Workbooks/Azure Web Application Firewall/WAF monitoring/WAF Monitoring.workbook
```diff
@@ -0,0 +1,831 @@
...
+        "size": 3,
+        "showAnalytics": true,
+        "title": "WAF actions filter",
+        "exportFieldName": "series",
+        "exportParameterName": "SelectedAction",
+        "exportDefaultValue": "*",
+        "showExportToExcel": true,
+        "exportToExcelOptions": "all",
+        "queryType": 0,
+        "resourceType": "microsoft.operationalinsights/workspaces",
```
gardnerjr, line 236: like some of the params, this query doesn't specify any workspace and isn't referenceing the parameter?  (i'm presuming you're testing this query inside workbooks ON a logs workspace, in that context it would inherit that workspace as the workspace to use. outside of logs (like on your resource type or in Azure Monitor -> Workbooks view) this template is all errors that you havn't selected   workspace:
image

sowmyam2019: I tested in Azure Monitor workbooks and I see the data displayed correctly 
image
monitoring metrics

Workbooks/Azure Web Application Firewall/WAF monitoring/WAF Monitoring.workbook
sowmyam2019: This is a workbook that is used by WAF team and customers already via the Sentinel ToC blade. We want to keep the workbook as-is since the original author is no longer with the company.

Workbooks/Azure Web Application Firewall/WAF monitoring/WAF Monitoring.workbook
sowmyam2019: This is a workbook that is used by WAF team and customers already via the Sentinel ToC blade. We want to keep the workbook as-is since the original author is no longer with the company.

Workbooks/Azure Web Application Firewall/WAF monitoring/WAF Monitoring.workbook
sowmyam2019: I have changed this to be Azure WAF Monitoring metrics

Workbooks/Azure Web Application Firewall/WAF monitoring/WAF Monitoring.workbook
```diff
@@ -0,0 +1,831 @@
...
+            "multiSelect": true,
+            "quote": "'",
+            "delimiter": ",",
+            "query": "let FakeData = (datatable (Message:string, ruleName_s:string, clientIp_s:string, clientIP_s:string, action_s:string, transactionId_s:string,trackingReference_s:string) [ \"\", \"\", \"\", \"\", \"\", \"\", \"\" ]);\r\nFakeData | union AzureDiagnostics\r\n| where ResourceType == \"APPLICATIONGATEWAYS\" or ResourceType == \"FRONTDOORS\" or ResourceType == \"PROFILES\" or ResourceType == \"CDNWEBAPPLICATIONFIREWALLPOLICIES\"\r\n| summarize Count=count() by ResourceType\r\n| extend ResourceTypeImproved = iif(ResourceType == \"APPLICATIONGATEWAYS\", \"Application Gateway\", ResourceType)\r\n| extend ResourceTypeImproved = iif(ResourceTypeImproved == \"FRONTDOORS\", \"Azure Front Door\", ResourceTypeImproved)\r\n| extend ResourceTypeImproved = iif(ResourceTypeImproved == \"PROFILES\", \"Azure Front Door Premium\", ResourceTypeImproved)\r\n| extend ResourceTypeImproved = iif(ResourceTypeImproved == \"CDNWEBAPPLICATIONFIREWALLPOLICIES\", \"Azure CDN\", ResourceTypeImproved)\r\n| order by Count desc, ResourceTypeImproved asc\r\n| project ResourceTypeImproved",
```
gardnerjr, line 145: if i'm interpreting this correctly, i think you could dramatically simplify this and every query below.

dropdowns support 2 columns of data:
1. a "value" (like "APPLICATIONGATEWAYS") generally used *in queries*
2. a "label" (like "Application Gateway") generally displayed *to users*

you could have this dropdown return 2 columns like that, and then when a user selects the "Application Gateway" item, the parameter value would be "APPLICATIONGATEWAYS"

(if there are fixed set of them that you know up front, you can list them a different way statically here and you don't even have to run a query at all (unless you want them to be filtered to only the existing values?)

if you do that, ALL of the queries below that have this big block:
```
| where (ResourceType == "APPLICATIONGATEWAYS" or ResourceType == "FRONTDOORS" or ResourceType == "PROFILES" or ResourceType == "CDNWEBAPPLICATIONFIREWALLPOLICIES") and ("{WAFType:label}" == "All" or (ResourceType == "APPLICATIONGATEWAYS" and "{WAFType:label}" contains "application gateway") or (ResourceType == "FRONTDOORS" and  "{WAFType:label}" contains "azure front door") or (ResourceType == "PROFILES" and "{WAFType:label}" contains "azure front door premium") or  (ResourceType=="CDNWEBAPPLICATIONFIREWALLPOLICIES" and "{WAFType:label}" contains "cdn")) and ("{WAF:label}" == "All" or Resource in ({WAF}))
```
(which alone is 613 ch in every query before any parameters get replaced)

to something like

```
let supportedTypes = dynamic(["APPLICATIONGATEWAYS",  "FRONTDOORS", "PROFILES", "CDNWEBAPPLICATIONFIREWALLPOLICIES"]);
let wafType = dynamic([{WAFType}]);
let wafResources = dynamic([{WAF}]);

///... other part of query...
| where ResourceType in (supportedTypes) // has_any might be even faster?
and ("*" in (wafType) or ResourceType in (wafType))
and ("*" in (wafResources ) or Resource in (wafResources))
///... other part of query

which is 344ch (without the comments, and before any parameters get replaced)

i believe this does the same thing much simpler and much more readable and easy to add to (the supported types could even be extracted out to a parameter and reused as well)

File: Workbooks/Azure Web Application Firewall/WAF monitoring/WAF Monitoring.workbook [hunk 11]
```diff
@@ -0,0 +1,831 @@
...
+            "type": 2,
+            "isRequired": true,
+            "multiSelect": true,
+            "quote": "'",
+            "delimiter": ",",
+            "query": "let FakeData = (datatable (Message:string, ruleName_s:string, clientIp_s:string, clientIP_s:string, action_s:string, transactionId_s:string,trackingReference_s:string) [ \"\", \"\", \"\", \"\", \"\", \"\", \"\" ]);\r\nFakeData | union AzureDiagnostics\r\n| where (ResourceType == \"APPLICATIONGATEWAYS\" or ResourceType == \"FRONTDOORS\" or ResourceType == \"PROFILES\" or ResourceType == \"CDNWEBAPPLICATIONFIREWALLPOLICIES\") and (\"{WAFType:label}\" == \"All\" or (ResourceType == \"APPLICATIONGATEWAYS\" and \"{WAFType:label}\" contains \"application gateway\") or (ResourceType == \"FRONTDOORS\" and  \"{WAFType:label}\" contains \"azure front door\") or  (ResourceType == \"PROFILES\" and \"{WAFType:label}\" contains \"azure front door premium\") or (ResourceType==\"CDNWEBAPPLICATIONFIREWALLPOLICIES\" and \"{WAFType:label}\" contains \"cdn\"))\r\n| summarize Count=count() by Resource\r\n| order by Count desc, Resource asc\r\n| project Value = Resource, Label = strcat(Resource, \" - \", Count)",
```
gardnerjr, line 175: a bunch of these queries are unioning with table with a fake row, which has different columns (including requestUri_s which isn't in my table? but is used in the filtering? is this all intentional?), and that blank row is *always* filtered out because of the way the query is written?

```
a 
| union b 
| where c
| summarize d
```

is *NOT* the same as
```
a 
| union (b 
| where c) // parentheses are important here!
|  summarize by d
```

if i add parens to some of these queries i get a single row of 0 count for a blank row.

so either these queries need parens to keep the union empty row intact, OR they don't need the union with fakedata at all

Workbooks/Azure Web Application Firewall/WAF monitoring/WAF Monitoring.workbook
gardnerjr: the word "Sandy" appears all over the place, in queries like 
```
| where policy_s == "sandycookieWAFpolicy" or policy_s  == "AmysAPIMwaf" or policy_s =="sandyappgwdcWAFpolicy"
```

in some queries it is commented out, in some it is not?

is sandy a feature name or something i don't know about or is this specific to your sample data and won't work anywhere else?

Workbooks/Azure Web Application Firewall/WAF monitoring/WAF Monitoring.workbook
tobystic: Are you referring to the default value for Time range ? if yes, then yes the last 24hours as the landing value makes sense.

Workbooks/Azure Web Application Firewall/WAF monitoring/WAF Monitoring.workbook
tobystic: Either the customer selects the workspace or all workspaces are set by default. In a workbook, customer can select the workspace, given the scope for which they are investigation or looking to focus. This could be set to All for a WAF pane workbook.

Workbooks/Azure Web Application Firewall/WAF monitoring/WAF Monitoring.workbook
tobystic: > like some of the params, this query doesn't specify any workspace and isn't referenceing the parameter? (i'm presuming you're testing this query inside workbooks ON a logs workspace, in that context it would inherit that workspace as the workspace to use. outside of logs (like on your resource type or in Azure Monitor -> Workbooks view) this template is all errors that you havn't selected workspace: image

Your assumption is correct "_testing this query inside workbooks ON a logs workspace, in that context it would inherit that workspace as the workspace to use_" . There is no "All" option here so it defaults to all.

sowmyam2019:
Closing this PR. Tobi will create a new one.

PR #2184, opened 2023-03-23
added pod name field to IP Subnet Usage Workbook

Khushbu-Parekh:
Added PodName field in IPs in use tab and removed time range filter

image


Workbooks/AKS/Subnet IP Usage/Subnet IP Usage.workbook
```diff
@@ -704,7 +641,7 @@
...
+              "query": "InsightsMetrics\n| where TimeGenerated < now() and TimeGenerated >= now(-120s)\n| where Name == \"cx_ipam_pod_allocated_ips\"\n| where Namespace == \"prometheus\"\n| extend jsonprop = parse_json(Tags)\n| extend HostName = jsonprop.hostName\n{hostwhereclause}\n| extend URL = tostring(jsonprop.scrapeUrl)\n| extend PodIP_trim_start = trim_start(\"http://\", URL)\n| extend PodIP = trim_end(\":10092/metrics\", PodIP_trim_start)\n| union KubePodInventory\n| project TimeGenerated, PodIP, HostName, Name\n| join (KubePodInventory\n  | project PodIp, Name, Namespace)\n  on $left.PodIP == $right.PodIp\n| extend PodName = Name \n| summarize arg_max(TimeGenerated, *) by PodIP, PodName, Namespace \n| project TimeGenerated, PodIP, PodName, HostName, Namespace \n",
```
vishiy, line 644: are time ranges intended to be hard coded (now() & now(-120s) ?

vishiy: ok. so are you limiting this to only last 2 minutes of data (120s based on your query) ? Would this help customers? i dont think anyone would react that quick to come and see the data here ?

Khushbu-Parekh: @vishwanath Confirmed we need only current time data

gardnerjr: (fyi, any sensitivity to logs ingestion delays?  2 minutes is not a lot of time to guarantee data might exist)

Khushbu-Parekh: @gardnerjr, how much time would you suggest to filter?

gardnerjr: the time range is entirely up to you, for your team to balance between how much data is useful and how much time prevents nothing from being found/displayed.  this is why most teams use time range parameters, with a short-ish default value but the ability for the user to change it for whatever reason (exploration, perf, diagnosing issues, etc)

gardnerjr (approved):
(unclear why our team is required, as we don't generally have the required resources to validate templates outside our team)

bragi92 (approved):

Issue #1982, opened 2022-07-29
Authenticated User Timeline Workbook - Parse Error


jackwagons:
The _Authenticated User Timeline_ Workbook does not escape `UserId` in Session Timeline sec.  

KQL snippet:

set query_take_max_records=10001;set truncationmaxsize=67108864;
let monthName = (month:int) {
    case(
        month == 1, 'January',
        month == 2, 'February',  
        month == 3, 'March',  
        month == 4, 'April',  
        month == 5, 'May',  
        month == 6, 'June',  
        month == 7, 'July',  
        month == 8, 'August',  
        month == 9, 'September',  
        month == 10, 'October',  
        month == 11, 'November',  
        month == 12, 'December',
        ''
    )
};
union customEvents, pageViews
| where timestamp > ago(1d)
| where user_AuthenticatedId == 'domain\testuser' **// needs to be escaped**
| where name in ('*') or '*' in ('*')
| summarize Activities = count(), Start = min(timestamp) by SessionId = session_Id
| order by Start desc
| project ['Session Start'] = strcat('⌚ ', monthName(monthofyear(Start)), ' ', dayofmonth(Start), ', ', tolower(format_datetime(Start, 'h:mm tt'))), Activities, SessionId
//'⌚'

Is the expectation that the source data is responsible for escaping?
If not, is there an accepted pattern or utility in other Workbooks? 

Happy to submit a PR based on feedback.

gardnerjr:
@mskailian i think this is an appinsights template, just needs to have that parameter use `:escape` in the query like `{parameterName:escape}` ?

PR #2222, opened 2023-04-10
UserId escape

MSKaiLian:
## PR Checklist
https://github.com/microsoft/Application-Insights-Workbooks/issues/1982
* escape UserId 
Before user_AuthenticatedId == 'domain	estuser'
After user_AuthenticatedId == 'domain\testuser'


MSKaiLian:
> are there any other similar templates using userid like this?

I don't see other userid like this

PR #2227, opened 2023-04-12
Adding advisor recommendations and log analytics integrations

StefArroyo:
## PR Checklist

Updating the SQL UI to have Advisor Recommendations and link to our new Log Analytics Workbooks/Troubleshooting workbooks.

image

* [ X ] ensure your template has a corresponding gallery entry in the gallery folder
* [ X ] ensure all steps have meaningful names
* [ X ] ensure all parameters and grid columns have display names set so they can be localized
* [ X ] ensure that parameters id values are unique __or they will fail PR validation__ (parameter ids are used for localization)
* [ X ] ensure that steps names are unique __or they will fail PR validation__ (step names are used for localization)
* [ X ] grep `/subscription/` and ensure that your parameters don't have any hardcoded resourceIds __or they will fail PR validation__
* [ X ] remove `fallbackResourceIds` and `fromTemplateId` fields from your template workbook __or they will fail PR validation__


erpope:
I just had some thoughts and I figured it is easier to put it here over marking in the workbook.
1. What are your thoughts on Adding Service Availability to the main Overview tab. The Availability tab only contains one chart and I think the availability metric itself is a good starting point when looking at Overview/Health of the customer's account.
2. For multi-region account, having Replication Latency might be a nice to have. It would be super cool if there can be a map with the customers regions and then a way to click and see the latency over time. 
3. 429 % would be nice :), we'll see about squeezing that work in
4. there are a lot of blank panels (with edit button) at the bottom of the workbook, what are those for our of curiosity?
5. I would move this section on the main overview to the top or the bottom, I think it separates the metrics which may make the customer miss the section at the bottom about Metrics by Collection
 
image

aliuy (approved):
LGTM!

PR #2234, opened 2023-04-19
Added new template for NSP Access logs insights

smurari90:
Objective: To show NSP(Network Security Perimeter) product access logs insights using workbook, so created new workbook template to show the access logs with time range parameter so that users can access the logs.

## PR Checklist

image

image

image

### DIFF
--- a/Workbooks/Network Security Perimeter/Access Logs/NSP Access Logs Insights.workbook
+++ b/Workbooks/Network Security Perimeter/Access Logs/NSP Access Logs Insights.workbook
@@ -1,0 +1,98 @@
+{
+  "version": "Notebook/1.0",
+  "items": [
+    {
+      "type": 9,
+      "content": {
+        "version": "KqlParameterItem/1.0",
+        "parameters": [
+          {
+            "id": "74ba6ba7-94ee-4a5d-9752-a643b2512995",
+            "version": "KqlParameterItem/1.0",
+            "name": "FilterResource",
+            "type": 1,
+            "isHiddenWhenLocked": true,
+            "timeContext": {
+              "durationMs": 2592000000
+            },
+            "value": ""
+          },
+          {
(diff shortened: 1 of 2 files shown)


arjun-d-patel (approved):

Workbooks/Network Security Perimeter/Access Logs/NSP Access Logs Insights.workbook
```diff
@@ -0,0 +1,98 @@
...
+        "version": "KqlItem/1.0",
+        "query": "NSPAccessLogs \r\n| where ServiceResourceId contains \"{FilterResource}\"",
+        "size": 0,
+        "showAnalytics": true,
+        "title": "Access Logs for the selected paas resource",
+        "timeContextFromParameter": "TimeRange",
+        "showExportToExcel": true,
+        "queryType": 0,
+        "resourceType": "microsoft.network/networksecurityperimeters",
+        "gridSettings": {
```
gardnerjr, line 90: all my suggestions would be nits: like having this query return only meaningful columns, with nice renderers/etc configured instead of just defaults/etc.

if it is the only thing in the page, you could also make the size larger to take up more space?

could also possibly optimize using `has` instead of `contains` in the query, but that's more up to exactly what you're doing in the query and if that's better/worse for you.

gardnerjr (approved):

Workbooks/Network Security Perimeter/Access Logs/NSP Access Logs Insights.workbook
smurari90: Good suggestions, I would have to get this page reviewed and take suggestions from PMs and users regarding columns so will update the column names in next update. I will look into has option more and update it later. Thanks @gardnerjr for all the support. Without your support this could not have been done.

PR #2310, opened 2023-05-31
Fixed BatchJob Tile Error, Changed Onboarding Link

jainayu-msft:
- Added a new query prompt that displays an error if user doesn't have atleast reader access to LA, else, shows that access is validated.
- Onboarding prompt now leads to a different page.
- Modified query for perf insights batch job tile such that it displays no results instead of error if batch jobs are not configured.
- Changed column header to "Action" instead of "Insights" in alert table.
- Changed column value to "Investigate" instead of "View" for consistency.

85e5d0eb-a811-40f9-8d98-4e90e5d71666

5c31d1ba-5cf5-43ad-ae15-935ee1e8222a

1f3f8748-7e56-488b-8c53-d8a0b4afb5f9

cb8fd65f-e8b8-4302-aacf-2796ebfbe9cc

7d727aff-59c0-4ac6-8ece-b9d487fd5f3b


jbaskar (approved):
Just add the list of bugs you have fixed in the description.

jbaskar (approved):

PR #2322, opened 2023-06-12
24185180: update prom explorer template in AMW to use this all string columns setting, high limit for max series

gardnerjr:
## Before
the chart just uses the default setting and shows 1 series:

image

## After
use all string columns, split to first 50+others

image


msft-alwag (approved):
Thanks!

PR #2328, opened 2023-06-16
NetWeaver Workbook - Add appropriate param check name for outbound queue tiles and updated logic to show sub menu for queue statistics tab

hsridharan:
## PR Checklist

* [x] Validate your changes using one or more of the [testing](../Documentation/Testing.md) methods.

Description:

1. This PR is a minor bug fix for outbound queue tiles. Ideally to show/hide the outbound queues we need to check the outbound queue parameter, but looks like the code was checking the "inbound queue parameter" and some customers who didn't have any data in outbound queues in the SAP system were seeing the below error in the workbook:
 
image

As a part of the fix, we updated the correct parameter to hide/show the tiles and also included proper step names for certain tiles to follow the appropriate nomenclature in the workbook.

IcM - https://portal.microsofticm.com/imp/v3/incidents/details/393660828/home

image

2. Sub menu for the Queue Statistics tab is currently not displayed today when a customer creates SOAP-only providers and does not have any RFC providers. This PR includes a fix to remove that check and show all three sub-menu items
image

Test case : https://microsoft-my.sharepoint.com/:x:/p/hsridharan/ESj-RgSbeghNqKAlLzGTM2MB4xpnUARWYyuOiPWlepM33A?e=lt4isX

### DIFF
--- a/Workbooks/SapMonitor2.0/SapNetWeaver/SapNetWeaver.workbook
+++ b/Workbooks/SapMonitor2.0/SapNetWeaver/SapNetWeaver.workbook
@@ -7928,7 +7928,7 @@
                     "comparison": "isEqualTo",
                     "value": "0"
                   },
-                  "name": "Total Inbound Queue Items - Copy"
+                  "name": "Total outbound Queue"
                 },
                 {
                   "type": 3,
@@ -7982,7 +7982,7 @@
                     "comparison": "isEqualTo",
                     "value": "0"
                   },
-                  "name": "Total Inbound Queue Items - Copy - Copy"
+                  "name": "Total outbound Queue Items"
                 },
                 {
                   "type": 3,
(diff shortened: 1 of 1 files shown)


devanshjainms:
I think the issue that the customer pointed was in the chart that we have which visualizes both the inbound and outbound queues.  The chart has a conditional visibility on the inbound queue, i.e., the chart is visible only when data is present in inbound queue.

image

image

hsridharan:
Added one additional check for the outbound table as well in this visualization because it uses both tables -  SapNetweaver_InboundQueues_CL and SapNetweaver_OutboundQueues_CL and the problem which is mentioned in the above comment is already addressed through previous PR - https://github.com/microsoft/Application-Insights-Workbooks/pull/2303

PR #2346, opened 2023-06-23
remove the hidden resource group param from keyvault insights

gardnerjr:
its hidden, always set to all
but it makes an extra unnecessary query to arg to get the RG's for all the subs and makes the queries more complicated

There's no *visible* or functional changes in insights, only when editing:

### before
image

you can see the "resource groups" all

### after
image

after, doesn't.


ericajlin:
Looks like at resource KV insights has the same issue - any plans on removing the parameter from there as well?

ericajlin (approved):

gardnerjr:
@ericajlin the at resource one is looking up sub/rg/etc?  it's already a single resource it shouldn't need any of that?  will look as a separate change?

PR #2351, opened 2023-06-28
Treat .workbook files as json in VSCode

dimitri-furman:
This PR associates .workbook files with the json formatter/syntax highlighter in VSCode for readability.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -4,5 +4,7 @@
         "MD025": {
             "front_matter_title": ""
         }
-    }
-}
+    },
+    "files.associations": {
+        "*.workbook": "json"
+    },}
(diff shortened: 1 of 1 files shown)


.vscode/settings.json
```diff
@@ -4,5 +4,7 @@
         "MD025": {
             "front_matter_title": ""
         }
-    }
-}
\ No newline at end of file
+    },
+    "files.associations": {
+        "*.workbook": "json"
+    },}
```
gardnerjr, line 10: is this valid json?  the preview/diff tool has a red thing here implying that it is not valid

dimitri-furman: Pushed the fix

gardnerjr (approved):

PR #2406, opened 2023-09-01
Show all ase alerts

vikramchandramohan:
## PR Checklist

* The ASE alerts tab currently does not show information about resolved alerts. This means that we cannot see information about, for example, an ASE restart from the previous day. We changed this to show all types of alerts from the ASE in this tab

Screenshot from before change:

Screenshot 2023-09-01 113348
image

Screenshot from after change:
image
image


awlodge:
Please can you include the alert condition in the columns shown, so the user can see if the alert is still active or is resolved? And then you can add some custom formatting to show an icon based on the alert condition, like in the main alerts blade:
image

Also, please apply the same change to the "all alerts" view on the overview tab.

PR #2413, opened 2023-09-06
Azure APIM Analytics workbook

VitaliyKurokhtin:
Adding new Analytics workbook we want to make available for all customers for Microsoft.ApiManagement/service resource.

## PR Checklist

* [ ] Validate your changes using one or more of the [testing](../Documentation/Testing.md) methods.


tomkerkhove (approved):

CODEOWNERS
```diff
@@ -172,8 +172,8 @@
 /gallery/cosmosdb-insights/** @microsoft/azurecosmosdb-workbooks
 /gallery/workbook/microsoft.azurepercept-accounts.json @microsoft/azure-percept-workbooks
 /gallery/workbook/microsoft.database-watcher.json @microsoft/sqlexternalmonitoring
-/gallery/apimanagement-service/** @microsoft/azure-api-management
 /gallery/workbook/microsoft.mobilenetwork-mobilenetwork-sites.json @microsoft/azure-private-mec-workbooks
+/gallery/workbook/apimanagement-service/microsoft.apimanagement-service.json @microsoft/azure-api-management
```
gardnerjr, line 176: you don't want this nested into a subdirectory, so this file should probably be `/gallery/workbook/rp.name-resource-type` like probably `/gallery/workbook/microsoft.apimanagement-service` if the resource type is `microsoft.apimanagement/service`

this would be a thing where validating this gallery actually works with the testing instructions would verify that what you pass to the workbook blades actually finds the correct gallery file or gets you a 404.

gardnerjr: actually it looks like the *real file* might be correctly named but this line here isn't?

gardnerjr (changes requested):
(i think you just need to remove `/apimanagement-service` from the new path listed in the CODEOWNERS file?)

gardnerjr (approved):

PR #2593, opened 2024-03-13
Dell APEX Multi Cluster workbook

paulbarry-dell:
Multi Cluster Workbook for Dell APEX

image

### DIFF
--- a/Workbooks/Azure Stack HCI/Dell/MultiCluster/Cluster.workbook
+++ b/Workbooks/Azure Stack HCI/Dell/MultiCluster/Cluster.workbook
@@ -6,7 +6,7 @@
       "content": {
         "version": "KqlParameterItem/1.0",
         "crossComponentResources": [
-          "{subscription}"
+          "value::all"
         ],
         "parameters": [
           {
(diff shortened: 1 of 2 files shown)


Workbooks/Azure Stack HCI/Dell/MultiCluster/Cluster.workbook
```diff
@@ -0,0 +1,1594 @@
...
+              "{subscription}"
+            ],
+            "isHiddenWhenLocked": true,
+            "timeContext": {
+              "durationMs": 0
+            },
+            "timeContextFromParameter": "timeRange",
+            "queryType": 1,
+            "resourceType": "microsoft.resourcegraph/resources",
+            "value": "2 monitored; 118 unmonitored"
```
sayaghosh58, line 192: Remove the value. This should be coming from the query.

sayaghosh58 (approved):

gardnerjr (approved):
approving for now because it is set to be preview only, but should probably have the common nosubscriptions behavior that other templates in azure monitor have, to better explain when this template can't work because the user doesn't have any appropriate resources.

PR #2615, opened 2024-04-09
Add trailing whitespace file path validation

ericajlin:
Github seems to allow trailing whitespaces when creating file paths using the Github UI. Ensure as part of validating template paths that there are no trailing whitespaces before or after slashes.

### DIFF
--- a/test/jsonValidationTest.js
+++ b/test/jsonValidationTest.js
@@ -26,6 +26,7 @@ const LocKeys = [
     "markDown" // specific to cohorts
 ];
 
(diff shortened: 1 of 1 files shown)


gardnerjr:
is trailing space the only way this could fail?  or does github let you create files with leading spaces too?

ericajlin:
> is trailing space the only way this could fail? or does github let you create files with leading spaces too?

regex includes checks for both trailing and leading slash. i can update the const name

gardnerjr (approved):

PR #2632, opened 2024-04-17
Added logic to handle no subscription case. Removed isPreview from Azure Monitor gallery entry

paulbarry-dell:
Update to Dell APEX Multi Cluster workbook:
- Added logic to handle no subscription case. 
- Removed isPreview from Azure Monitor gallery entry

image

image


sayaghosh58 (approved):

gallery/workbook/Azure Monitor.json
```diff
@@ -105,8 +105,7 @@
 					"id": "Workbooks/Azure Stack HCI/Dell/MultiCluster",
 					"author": "Dell",
 					"description": "Multi Cluster Event Monitoring for Dell APEX Cloud Platform for Microsoft Azure",
-					"name": "Multi Cluster Event Monitoring for Dell APEX Cloud Platform for Microsoft Azure",
-					"isPreview": true
+					"name": "Multi Cluster Event Monitoring for Dell APEX Cloud Platform for Microsoft Azure"
```
gardnerjr, line 108: why is the name the same as the description?  the name should be something short so it isn't "multi cluster even....." and then the description has the same content?

paulbarry-dell: Thanks John, will review this with documentation

gardnerjr (approved):

PR #2684, opened 2024-06-17
Scale disk graph in IoT Edge device details.

alaendle:
This solves the problem of to fine granular disk space usage graph in the azure edge hub device details workbook.

Since the amount of data is dynamic (it depends on the number of disks and not only on the sampling frequency) the current strategy of bining just to the sampling frequency can lead to "too many points" errors.

grafik

This can be avoided by choosing the `max_of({timeRange:grain}, time(\"{samplingFrequency}\"))` and not only the `samplingFrequency`. 

The `maxOf` avoids the generation of too fine grained bins (smaller than the sampling rate) and the `grain` increases the bin-size if large time frames are used. The data isn't distorted since the max-value is chosen for aggregation - so huger bin's do no harm.

Please let me know if I was unclear or you need further details. Thanks in advance for considering this change.


alaendle:
Just to add the math behind it - at a sampling rate of 5 minutes you generate around ~ 2000 data-points per week. With logical volumes we have usually have 7 "disks" - so without further scaling we end up with ~ 14000 data-points for seven days - giving the mentioned error.

damonbarry (approved):

PR #2807, opened 2024-10-14
Mpbastos/avdiupdate

mpbastos:
- Add case insensitive regex on the AssignedDCR parameter to select the default DCR.
- Upate poolla to use ARG query on all subscriptions instead of selected.

[Private Branch](https://portal.azure.com/?feature.workbookGalleryBranch=mpbastos/avdiupdate)


xiangtin (approved):

xiangtin:
Need to update the version number.

mike-mo (changes requested):
Need version number update

mike-mo (approved):

PR #2814, opened 2024-10-18
Updated Azure Services Retirement workbook

v-maytri0802:
Added the below info banner

image


v-maytri0802:
stuck at the license/ca step

PR #2846, opened 2024-11-26
Removed extra folder in path

mepples21:
## Summary

I removed an extra folder in the path to the workbook to ensure it doesn't interfere with the build process.

## Screenshots

- [x] If you added a template to a gallery, show a screenshot of it in the gallery view (which verifies its shows up where you expected).

## Validation

- [x] Validate your changes using one or more of the [testing](https://github.com/microsoft/Application-Insights-Workbooks/blob/master/Documentation/Testing.md) methods.
  
## Checklist

- [x] If you are adding a new template, gallery, or folder, add your team and folder/file(s) to the CODEOWNERS file at the root of the repo. CODEOWNERS entries should be teams, not individuals.
- [x] Ensure all steps in your template have meaningful names.
- [x] Ensure all parameters and grid columns have display names set so they can be localized.


mepples21:
@JefTek could you please take a look at this PR?

JefTek (approved):
I see the update to the path

PR #2873, opened 2025-01-16
Remove sql insights from workbook gallery

masree:
## Summary

SQL Insights officially got retired on 31 December 2024. Removing its reference from workbook insights gallery

## Screenshots

- [x] If you added a template to a gallery, show a screenshot of it in the gallery view (which verifies its shows up where you expected).
image

## Validation

- [x] Validate your changes using one or more of the [testing](https://github.com/microsoft/Application-Insights-Workbooks/blob/master/Documentation/Testing.md) methods.
Validated, no SQL Insights under "Insights" gallery
  
## Checklist

### DIFF
--- a/gallery/workload-insights/Azure Monitor.json
+++ b/gallery/workload-insights/Azure Monitor.json
@@ -15,12 +15,6 @@
 			]
 		},
 		{
-			"id": "SQL",
-			"name": "SQL (preview)",
-			"templates": [
-			]
-		},
-		{
 			"id": "azureMonitorInsights",
 			"name": "Insights",
 			"templates": [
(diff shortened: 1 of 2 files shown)


gallery/workload-insights/Azure Monitor.json
```diff
@@ -18,42 +18,12 @@
 			"id": "SQL",
```
dimitri-furman, line 18: Remove the entire object with `id` `SQL`

masree: Yes I thought about it, I was not sure if we can remove it wholesale as called out by John. So left it as-is. If you think its ok to remove, I can

gardnerjr (approved):

dimitri-furman (approved):

PR #2892, opened 2025-02-05
Adding a tab for Data Collection Rules

noakup:
Adding a tab for Data Collection Rules streaming to this workspace, showing their details and if a system DCE is set to this workspace, and warnings if they don't use this DCE.


osalzberg:
#sign-off

osalzberg (approved):

PR #2900, opened 2025-02-12
Update contributing to be more explicit about teams in github

gardnerjr:
## Summary

Update contributing to be more explicit about teams in github

No changes to templates or any other content.


matteing (approved):
Looks good.

PR #2939, opened 2025-03-10
Call Insights - Fix timeseries chart title and fix query to remove "success" logs in table

bluk16:
1. Remove `(Bottom legend shows reliability when hover, and it shows average reliability of the series when unhover)`
from table title
image

2. Fix query to not show "success" logs in top issue table
[Local test URL](https://portal.azure.com/?feature.workbookGalleryBranch=brianluk/fix_reliability_UI_query)


dvillasenor-msft (approved):
Verified the text is updated, and I no longer see the 'Success' metric for APIs.

PR #2990, opened 2025-04-17
Address some filters bugs and add malicious IP details in overview page

srisakthimaheswari:
## Summary
When both flowtypes are selected, and only one of the flowtypes' schema is present, we do not show any data and do not surface up the right error messages. This PR corrects that behaviour and we surface up the right warning and show the data for the schema that we could find.
It also handles scenarios:
- When workspace is deleted
- When no schema is present
- When only topology schema is present
- When filters are not set right
- When only one type of schema is present
- When both schema is present but only 1 has data
- When schema is present but no data

The Top Blocked NSG/ACL section will not include the Malicious IP to which there is traffic than just the azure resource. 
## Screenshots

Before:
image

After:
image

Malicious IP details in grid:
image

## Validation

Link: https://ms.portal.azure.com/?feature.workbookGalleryBranch=srisa%2FWBBugs#view/Microsoft_Azure_Network/NetworkWatcherMenuBlade/~/trafficAnalytics
image

## Checklist

- [X] If you are adding a new template, gallery, or folder, add your team and folder/file(s) to the CODEOWNERS file at the root of the repo. CODEOWNERS entries should be teams, not individuals.
- [X] Ensure all steps in your template have meaningful names.
- [X] Ensure all parameters and grid columns have display names set so they can be localized.


samitabh-ms (approved):
LGTM
[Approving to meet review requirements]

PR #3044, opened 2025-06-05
Modify minimum aggregation to match pg_qs.interval_length_minutes

seernu:
## Summary

Query Performance insights will change the minimum aggregation to match the GUC (server parameter) pg_qs.interval_length_minutes - which is the interval we capture the data.

this interval will only be used for time windows less then 12 hours

## Screenshots
OLD:
image

NEW:
image

## Validation

- [ ] Validate your changes using one or more of the [testing](https://github.com/microsoft/Application-Insights-Workbooks/blob/master/Documentation/Testing.md) methods.
  
## Checklist


nachoalonsoportillo (approved):
LGTM

PR #3069, opened 2025-07-28
Fixing TimeRange in Insights Blade

patty-chow:
Solves customer IcM where they couldn't edit their metrics by time range

## Summary

There was a metrics display issue within the Insights dashboard where, despite metrics data flowing correctly, the dashboard fails to expand the selected time range beyond one hour.

## Screenshots

Before fix (from IcM):
image

After fix (different dataset):
image

## Validation

Validated using Advanced Editor

## Checklist

- [x] If you are adding a new template, gallery, or folder, add your team and folder/file(s) to the CODEOWNERS file at the root of the repo. CODEOWNERS entries should be teams, not individuals.
- [x] Ensure all steps in your template have meaningful names.
- [x] Ensure all parameters and grid columns have display names set so they can be localized.

### DIFF
--- a/Workbooks/CosmosDb Mongo vCore/Overview/Overview.workbook
+++ b/Workbooks/CosmosDb Mongo vCore/Overview/Overview.workbook
@@ -13,7 +13,6 @@
             "label": "Azure Cosmos DB",
             "type": 5,
             "isRequired": true,
-            "isGlobal": true,
             "isHiddenWhenLocked": true,
             "typeSettings": {
               "additionalResourceOptions": [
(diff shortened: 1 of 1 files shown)


Workbooks/CosmosDb Mongo vCore/Overview/Overview.workbook
```diff
@@ -13,13 +13,15 @@
             "label": "Azure Cosmos DB",
             "type": 5,
             "isRequired": true,
-            "value": "value::1",
+            "isGlobal": true,
```
gardnerjr, line 16: this change to make this hidden resource param global probably isn't correct or necessary?  nothing else "downstream" in the template modifies this resource param, does it?

gardnerjr (changes requested):

PR #3221, opened 2026-08-14
Add advisory PR workflow to analyze changed workbook templates

gardnerjr:
## Summary

Reviewers currently have no automated feedback on the best-practice analyzer (`scripts/src/AnalyzeTemplates.ts`) when a PR touches template content. This adds a new `pull_request`-triggered GitHub Action that runs that analyzer against the templates changed in a PR and surfaces its warnings/errors as a comment, so authors see issues (missing localization labels, etc.) before merge.

The workflow is deliberately **advisory only**: it never fails the build and does not block the PR. It:

- Builds the analyzer (`scripts` -> `tsc`).
- Diffs the PR to find changed `.workbook`/`.json` files (skips deleted files).
- Runs `AnalyzeTemplates.js --severity:2` on just those files.
- Posts/updates a single sticky comment with the findings in a collapsible block.

It is a separate file (`.github/workflows/template-pr-analysis.yaml`); the existing `template-validation.yaml` is untouched.

A couple of non-obvious details worth a look during review:

- **Findings are counted from analyzer output, not the exit code.** The analyzer prints findings to stderr and only sets exit code 1 for errors/critical, so warning-only runs still exit 0. The comment keys off a count of finding lines to avoid falsely reporting "no issues."
- **Changed files are passed individually, not as folders.** The analyzer's folder-recursion path joins with a Windows `\\` separator, which would break on `ubuntu-latest`. Passing files directly avoids that code path.
- **`continue-on-error` on the comment step** keeps the job green on fork PRs, where `GITHUB_TOKEN` is read-only and commenting would otherwise fail.

Severity is tunable via `--severity:1` (errors only) or `--ignore:RULE` to mute specific rules.

## Screenshots

N/A - CI workflow change, no template or gallery content added.

## Validation

Verified locally: `npm ci` and `npm run build` in `scripts` succeed, and `node dist/AnalyzeTemplates.js --severity:2` against a real workbook produces the expected `PARAM001` warning. Confirmed the finding-count regex matches the analyzer's output format.

## Checklist

- [x] N/A - no new template, gallery, or folder added, so no CODEOWNERS change needed.
- [x] N/A - no template steps added.
- [x] N/A - no parameters or grid columns added.

### DIFF
--- a/.github/workflows/template-pr-analysis.yaml
+++ b/.github/workflows/template-pr-analysis.yaml
@@ -15,12 +15,12 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4.4.0
         with:
           fetch-depth: 0
 
       - name: Use Node.js
-        uses: actions/setup-node@v4
+        uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
         with:
           node-version: 24.x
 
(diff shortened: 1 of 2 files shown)


gardnerjr:
@copilot resolve the merge conflicts in this pull request

gardnerjr:
@copilot similar to https://github.com/microsoft/Application-Insights-Workbooks/pull/3228, this needs to have the actions updated to explicit versions?

PR #3209, opened 2026-07-13
Migrate SQL VA workbook queries from sub-assessments to assessments

omerbenishay12:
## Summary

Migrates the 9 **SQL Vulnerability Assessment** queries in the MDC `Vulnerabilities` workbook from the deprecated **sub-assessments** model (`microsoft.security/assessments/subassessments`) and the legacy **aggregate-assessment** model to the new **scanner-based assessments** model:

```
securityresources
| where type =~ "microsoft.security/assessments"
| extend scanner = tostring(coalesce(properties.additionalData.Scanner, properties.additionalData.scanner))
| where scanner =~ "SQL Vulnerability Assessment"
```

The SQL VA data source has moved from sub-assessments to first-class assessments. These queries were still reading the old sub-assessment resources, which are being deprecated. This updates them to read the equivalent data from the new scanner-based assessments while preserving each query's output columns so the grids/visualizations render unchanged.

## Queries updated

| # | Tile | Change |
|---|------|--------|
| 1 | Unhealthy SQL resources by type | aggregate-assessment -> scanner assessments; count distinct parent resources |
| 2 | Distinct SQL vulnerabilities by severity | sub-assessments -> scanner assessments |
| 3 | Vulnerabilities by severity per SQL resource | sub-assessments -> scanner assessments |
| 4 | Vulnerable SQL resources (selector) | sub-assessments -> scanner assessments |
| 5 | Vulnerable databases per server | sub-assessments -> scanner assessments |
| 6 | Grouped by benchmark | sub-assessments -> scanner assessments |
| 7 | Grouped by resource | sub-assessments -> scanner assessments |
| 8 | Grouped by severity | sub-assessments -> scanner assessments |
| 9 | Grouped by category | sub-assessments -> scanner assessments |

### Field mapping
- `VulnId`: `properties.id` -> `additionalData.ruleId`
- `Description`: `properties.displayName` -> `properties.metadata.displayName`
- `Severity` / `Category` / `Impact` / `Remediation` / `Query` / `Benchmarks`: `properties.*` -> `additionalData.*`
- `TimeGenerated`: `properties.timeGenerated` -> `properties.status.statusChangeDate`
- Parent server: extracted-from-id -> `additionalData.parentResource`
- Database: `resourceDetails.id` (+ `securityentitydata/sqlservers:` -> `/servers/.../databases/...` normalization for SQL-on-machine)

Also adds an `isnotempty(ParentResource)` guard and case-insensitive resource-group extraction.

[...]

PR #3222, opened 2026-08-16
Migrate machine and container VA queries from sub-assessments to assessments

LiorElb:
Follow-up to #3209, which migrated the SQL VA queries off sub-assessments. This does the same for the **machine (server)** and **container** tiles, which are currently broken for customers.

## Problem

Two independent regressions:

**1. Machine tiles render empty.** The 6 machine queries filtered `securityresources` on a hardcoded Qualys assessment key against `microsoft.security/assessments/subassessments`. That table now returns **0 rows tenant-wide**, so every machine tile is blank.

**2. CVE severity shows as `Unknown`.** CVE metadata is no longer emitted inline on the assessment record. On the tenant I validated against, inline severity is `"Unknown"` for **2064 of 2105** machine CVEs (~98%). Containers are less affected but still lose ~7%.

### 1. Machine queries → `microsoft.security/assessments`

Replaced the hardcoded assessment key with a category + resource-type filter:

```kusto
| where type == "microsoft.security/assessments"
| where properties.metadata.recommendationCategory == "SoftwareUpdate"
| where tostring(properties.resourceDetails.ResourceType) == "Microsoft.Compute/virtualMachines"
     or tostring(properties.resourceDetails.ResourceType) == "Microsoft.HybridCompute/machines"
| where tostring(properties.status.code) == 'Unhealthy'
```

The `Unhealthy` filter is required: the old `subassessments` table contained only findings, whereas `assessments` also holds healthy rows, which would otherwise inflate every count.

One related fix: `items[16]` grouped by `properties.category`, which does not exist on the new table. It now groups by `PackageType`.

### 2. CVE metadata → `microsoft.security/cvedetails`

Severity, description, remediation and impact now come from a `leftouter` join, for **both** machines and containers:

```kusto
| join kind=leftouter (
    securityresources
    | where type =~ "microsoft.security/cvedetails"
    | project cveId = tolower(tostring(properties.cveId)), cveSeverity = tostring(properties.severity)
) on cveId
```

For containers the inline value is retained as a fallback (join value wins → inline → `'Unknown'`). This resolves 4,998 of 5,031 previously-`Unknown` container CVEs (99.3%).

> `microsoft.security/cvedetails` is currently undocumented. The schema used here was derived empirically: 379,122 rows, exactly one row per `cveId`, keys are lowercase-first (`properties.cveId`, `properties.severity`). Note `severity`, not `Severity` — KQL dynamic access is case-sensitive and the capitalised form silently returns null.

### 3. Tenant scope (required by #2)

`cvedetails` rows are tenant-level and carry **no `subscriptionId`** (`id` = `/providers/microsoft.security/cvedetails/cve-2006-1451`). A subscription-scoped query returns nothing, so all 21 ARG queries move to:

```json
```

**The UI is unchanged.** The subscription picker is preserved and now applies as a KQL filter, injected as line 2 of every query:

```kusto
| where '*' in ({Subscription}) or subscriptionId in~ ({Subscription:subid})
```

The picker gains `"selectAllValue": "*"` so that **All** emits a single sentinel rather than expanding to every subscription ID. Without this, the longest query rendered at 33,493 characters against ARG's ~16,000 limit and every tile errored on load in large tenants; with it, 2,296 characters.

The guard is deliberately **not** applied inside the `cvedetails` sub-query — those rows have no `subscriptionId` and would all be filtered out.

## Consequences worth reviewing

**Critical severity column.** `cvedetails` emits a `Critical` severity that the workbook never rendered. Without a `sevC` column those CVEs count toward `Total` but land in no bucket. Added to the per-resource grids, pie chart and severity filters. Downstream of this, the container grid's `| top 500 by sevH desc` became `| order by ... | take 500`, because ARG's `top` operator accepts only a single sort key.

**Behavioural note.** "All" now means tenant-wide RBAC-visible scope, which may be broader than the previous subscription-list scope.

## Additional bug fix — happy to split out

The container queries now deduplicate severity before the join:

```kusto
| summarize inlineSev = min(inlineSev) by <keys>, cveId
```

This is **pre-existing and unrelated to the regression**. The same CVE can carry different inline severities across images in a registry, so `dcountif` counted it in multiple buckets. On one registry:

```
Total (dcount)  4855
sevC 220 + sevH 1874 + sevM 1705 + sevL 290 + sevU 3232  =  7321   (151% of Total)
```

After the fix the buckets reconcile (Total 4833; C 280 + H 2200 + M 1921 + L 304). It also keeps the join cheap by collapsing the left side from ~8.4M rows to ~5K distinct CVEs. The machine queries do not have this bug (`cvedetails` is strictly one row per CVE).

If reviewers would rather keep this PR to the regression only, I'm glad to pull it into a separate PR.

## Validation

All 21 queries were executed live against a production tenant via the ARG REST API at tenant scope — **0 failures**. The 6 container queries were additionally smoke-tested for row counts and column shape.

Note that `az graph query --management-groups <root>` does **not** give true tenant scope and returns 0 rows for `cvedetails`; the REST API with no scope specified is required.

## Screenshot

I don't have portal access to this template under a tenant with representative data, so I wasn't able to attach a rendered screenshot per CONTRIBUTING.md. Happy to add one if a reviewer can point me at a suitable tenant, or if a member of @microsoft/defender-for-cloud-workbooks can grab one.

### DIFF
--- a/Workbooks/Azure Security Center/Vulnerabilities/Vulnerabilities.workbook
+++ b/Workbooks/Azure Security Center/Vulnerabilities/Vulnerabilities.workbook
@@ -27,7 +27,8 @@
                 "value::all"
               ],
               "includeAll": false,
-              "showDefault": false
+              "showDefault": false,
+              "selectAllValue": "*"
             },
             "timeContext": {
               "durationMs": 86400000
@@ -129,14 +130,14 @@
       "type": 3,
       "content": {
         "version": "KqlItem/1.0",
-        "query": "securityresources\r\n| where type == \"microsoft.security/assessments/subassessments\"\r\n| extend assessmentKey = extract(\".*assessments/(.+?)/.*\",1,  id)\r\n| where assessmentKey == \"1195afff-c881-495e-9bc5-1486211ae03f\"\r\n| project ResourceType=tostring(split(id,\"/\").[6]), ResourceName=tostring(properties.resourceDetails.id)\r\n| summarize Count=dcount(ResourceName) by ResourceType",
+        "query": "securityresources\r\n| where '*' in ({Subscription}) or subscriptionId in~ ({Subscription:subid})\r\n| where type == \"microsoft.security/assessments\"\r\n| where properties.metadata.recommendationCategory == \"SoftwareUpdate\"\r\n| where tostring(properties.resourceDetails.ResourceType) == \"Microsoft.Compute/virtualMachines\" or tostring(properties.resourceDetails.ResourceType) == \"Microsoft.HybridCompute/machines\"\r\n| where tostring(properties.status.code) == 'Unhealthy'\r\n| project ResourceType = tostring(properties.resourceDetails.ResourceType), ResourceName = tolower(tostring(properties.resourceDetails.Id))\r\n| summarize by ResourceType, ResourceName\r\n| summarize Count = count() by ResourceType",
         "size": 4,
         "title": "Unhealthy machines",
         "noDataMessage": "No unhealthy machines found",
         "queryType": 1,
-        "resourceType": "microsoft.resourcegraph/resources",
+        "resourceType": "microsoft.resources/tenants",
         "crossComponentResources": [
-          "{Subscription}"
+          "value::tenant"
         ],
         "visualization": "table",
         "gridSettings": {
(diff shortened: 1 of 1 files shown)


LiorElb:
Closing in favour of a split, as requested in review.

> **@AdiKlinger:** I recommend not mixing VA related changes with Container changes. In case of regression in Containers, we don't want to need to rollback the whole commit.

> **@EladKamin:** please split the changes.

> **@HadarLivni:** So can you split just the VA changes with the scope changes on all queries, without changing the containers queries?

This is now:

- **#3223** — machine/VA tiles: `subassessments` → `assessments`, plus CVE severity recovery
- **#3224** — container tiles: CVE severity recovery, plus a fix for pre-existing bucket double-counting

**The two are fully independent.** They touch **zero overlapping JSON leaves**, are both branched from `master`, merge cleanly in either order, and either can be reverted alone without affecting the other. There is no ordering dependency.

Two changes since this PR:

1. **`selectAllValue` is gone.** I had assumed a single parameter could not serve both `crossComponentResources` and a `{Subscription:subid}` KQL guard under **All**. I verified this in the portal instead of assuming, and it was wrong — Workbooks expands per format specifier, so both mechanisms work off the unmodified parameter simultaneously. A tenant-scoped and a subscription-scoped count card both returned 333 across a 333-subscription tenant. The `Subscription` parameter is now **untouched** by both PRs, and the guard simplifies to `| where subscriptionId in~ ({Subscription:subid})`.

2. **SQL tiles are left alone.** They reference neither `CvesDetails` nor `assessmentKey`, so they are unaffected by both customer issues and did not need rescoping.

Net effect: this PR was +98/−64. The split is **+51/−18** and **+18/−18**.

PR #3223, opened 2026-08-17
Fix machine VA tiles: migrate to assessments and recover CVE severity

LiorElb:
## Summary

Fixes the **machine (server)** vulnerability assessment tiles in the Vulnerabilities workbook.

This replaces #3222, which reviewers asked be split so a container regression could be rolled back without reverting the machine fixes. Container changes ship separately in #3224 — **the two are independent, have no ordering dependency, and either can be reverted alone.**

## The customer issues

**1. `assessmentKey` filtering on Qualys sub-assessments stopped returning data.** Every machine tile renders empty.

**2. CVE details are no longer emitted inline** in `properties.additionalData.CvesDetails`. Severity breakdowns collapse into "Unknown" — currently **92.7%** of server rows (4,039 of 4,356).

## The fixes

| # | Change | Why |
|---|---|---|
| 1 | Rewrite the six machine queries from `microsoft.security/subassessments` to `microsoft.security/assessments` | Customer issue 1 |
| 2 | Join `microsoft.security/cvedetails` to recover CVE severity | Customer issue 2 |
| 3 | Move affected queries to tenant scope | Forced by 2 — `cvedetails` only resolves at tenant scope |
| 4 | Add a KQL subscription guard | Preserves the existing subscription picker |
| 5 | Add a Critical severity column | Enabled by 2 |

### Tenant scope + subscription filter

`microsoft.security/cvedetails` returns rows only at tenant scope, so the affected queries change to:

```jsonc
```

To keep the subscription picker behaving exactly as before, each rescoped query gets a guard as line 2:

```kql
| where subscriptionId in~ ({Subscription:subid})
```

**The `Subscription` parameter itself is unchanged.** No `selectAllValue`, no sentinel. This was verified in the portal rather than assumed — under **All**, Workbooks expands per format specifier:

- `{Subscription}` → `'/subscriptions/<guid>', …` (feeds `crossComponentResources`)
- `{Subscription:subid}` → `'<guid>', …` (feeds the KQL guard)

Both mechanisms agree: a tenant-scoped count card and a subscription-scoped count card both returned **333** across a 333-subscription tenant.

## Notes on `cvedetails`

- Rows have **no `subscriptionId`** (`id` is `/providers/microsoft.security/cvedetails/cve-2006-1451`), so the guard is deliberately **not** injected into the join subquery.
- The field is `severity` (lowercase-first); capital `Severity` silently returns null.
- 379,122 rows, exactly one per `cveId`. `name` is lowercase and `properties.cveId` uppercase, so the join is on `tolower()` both sides.

## Scope

- **Touched:** machine/VA items only — `items[6, 9, 12, 15, 16, 17]`
- **Untouched:** container items, all SQL items, and the shared `Subscription` parameter
- Zero overlapping JSON leaves with #3224 — verified

## Validation

- All 6 changed queries executed live against a 333-subscription tenant: **6/6 HTTP 200 with real data**
- Three drill-down tiles return 0 rows only because they are gated on an unsubstituted `{selectedServer}`; re-run with `All` they return data correctly
- Structural validation: valid JSON, 25 items, guard on line 2 of all 6, exactly one `subscriptionId in~` per query (none leaked into join subqueries)

## Known limitation (pre-existing, not introduced here)

`items[16]` ("Vulnerabilities by category") still `mv-expand`s `properties.additionalData.CvesDetails` for per-CVE fan-out. It groups by `PackageType` rather than severity, so it does not need the `cvedetails` join today — but if inline CVE data disappears entirely, that tile will drop to zero. Happy to address in this PR or as a follow-up, whichever reviewers prefer.

Closes #3222


LiorElb:
@microsoft-github-policy-service agree company="Microsoft"

hadarlivni (approved):

ahmedmasalha (approved):

PR #3224, opened 2026-08-17
Fix container VA tiles: recover CVE severity and correct bucket double-counting

LiorElb:
## Summary

Fixes the **container registry** vulnerability assessment tiles in the Vulnerabilities workbook.

This is the container half of #3222, which reviewers asked be split so a container regression could be rolled back without reverting the machine fixes. The machine half is #3223 — **the two are independent, have no ordering dependency, and either can be reverted alone.**

## The problems

**1. CVE details are no longer emitted inline** in `properties.additionalData.CvesDetails`. **54.0%** of container rows (4,826,322 of 8,933,263) now carry no usable severity. Rendered tenant-wide across 28 container resources, **27.5%** of findings land outside every severity bucket (5,664 of 20,586).

**2. The existing `dcountif` shape double-counts — this is a pre-existing bug.** A CVE carrying more than one inline severity is counted once per severity, so the buckets do not reconcile against the total:

| | Original |
|---|---|
| Reported total | 4,855 |
| Sum of severity buckets | 220 + 1,874 + 1,705 + 290 + 3,232 = **7,321** |
| | **151% of total** |

## The fixes

| # | Change | Why |
|---|---|---|
| 1 | Join `microsoft.security/cvedetails` to recover CVE severity | Problem 1 |
| 2 | Deduplicate before the join so each CVE contributes once | Problem 2 |
| 3 | Move affected queries to tenant scope | Forced by 1 — `cvedetails` only resolves at tenant scope |
| 4 | Add a KQL subscription guard | Preserves the existing subscription picker |

### Result

Of the 1,959 distinct container CVEs with no usable inline severity, the `cvedetails` join rescues **1,867 (95%)**, lifting overall coverage to **87%** (6,760 of 7,750). Unknown drops from 3,232 to roughly 128.

Buckets now reconcile: Total 4,833 — Critical 280, High 2,200, Medium 1,921, Low 304.

### Tenant scope + subscription filter

`microsoft.security/cvedetails` returns rows only at tenant scope, so the affected queries change to:

```jsonc
```

To keep the subscription picker behaving exactly as before, each rescoped query gets a guard as line 2:

```kql
| where subscriptionId in~ ({Subscription:subid})
```

**The `Subscription` parameter itself is unchanged.** No `selectAllValue`, no sentinel. This was verified in the portal rather than assumed — under **All**, Workbooks expands per format specifier:

- `{Subscription}` → `'/subscriptions/<guid>', …` (feeds `crossComponentResources`)
- `{Subscription:subid}` → `'<guid>', …` (feeds the KQL guard)

Both mechanisms agree: a tenant-scoped count card and a subscription-scoped count card both returned **333** across a 333-subscription tenant.

## Notes on `cvedetails`

- Rows have **no `subscriptionId`** (`id` is `/providers/microsoft.security/cvedetails/cve-2006-1451`), so the guard is deliberately **not** injected into the join subquery.
- The field is `severity` (lowercase-first); capital `Severity` silently returns null.
- 379,122 rows, exactly one per `cveId`. `name` is lowercase and `properties.cveId` uppercase, so the join is on `tolower()` both sides.

## Scope

- **Touched:** container items only — `items[7, 10, 13, 18, 19, 20]`. Query and scope changes only; no `gridSettings` changes.
- **Untouched:** machine items, all SQL items, and the shared `Subscription` parameter
- Zero overlapping JSON leaves with #3223 — verified

## Validation

- All 6 changed queries executed live against a 333-subscription tenant: **6/6 HTTP 200 with real data**
- Two drill-down tiles return 0 rows only because they are gated on an unsubstituted `{selectedACR}`; re-run with `All` they return data correctly
- Structural validation: valid JSON, 25 items, guard on line 2 of all 6, exactly one `subscriptionId in~` per query (none leaked into join subqueries)
- Branches merge cleanly with #3223 in either order

Part of #3222

### DIFF
--- a/Workbooks/Azure Security Center/Vulnerabilities/Vulnerabilities.workbook
+++ b/Workbooks/Azure Security Center/Vulnerabilities/Vulnerabilities.workbook
@@ -237,14 +237,14 @@
       "type": 3,
       "content": {
         "version": "KqlItem/1.0",
-        "query": "securityresources\n| where type == \"microsoft.security/assessments\"\n| where properties.metadata.recommendationCategory == \"SoftwareUpdate\"\n| where properties.resourceDetails.ResourceType == \".containerimage\"\n| where properties.resourceDetails.Source == \"Azure\"\n| extend status = tostring(properties.status.code)\n| where status == \"Unhealthy\"\n| extend Resource = tolower(extract(@'(?i)(.*?)/securityentitydata/([^/]+)', 1, id))\n| extend ResourceType = \"microsoft.containerregistry/registries\"\n| summarize Count=dcount(Resource) by ResourceType",
+        "query": "securityresources\n| where subscriptionId in~ ({Subscription:subid})\n| where type == \"microsoft.security/assessments\"\n| where properties.metadata.recommendationCategory == \"SoftwareUpdate\"\n| where properties.resourceDetails.ResourceType == \".containerimage\"\n| where properties.resourceDetails.Source == \"Azure\"\n| where tostring(properties.status.code) == \"Unhealthy\"\n| extend Resource = tolower(extract(@'(?i)(.*?)/securityentitydata/([^/]+)', 1, id))\n| extend ResourceType = \"microsoft.containerregistry/registries\"\n| summarize by Resource, ResourceType\n| summarize Count = count() by ResourceType",
         "size": 4,
         "title": "Unhealthy container registries",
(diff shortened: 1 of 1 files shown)


LiorElb:
@microsoft-github-policy-service agree company="Microsoft"

liorarviv (approved):

ahmedmasalha (approved):

PR #3232, opened 2026-09-02
Iriaosara io add throttlingworkbook

iriaosara:
## Summary

Explain your changes, so people looking at the PR know *what* and *why*, the code changes are the *how*.

## Screenshots

- [ ] If you added a template to a gallery, show a screenshot of it in the gallery view (which verifies its shows up where you expected).

## Validation

- [ ] Validate your changes using one or more of the [testing](https://github.com/microsoft/Application-Insights-Workbooks/blob/master/Documentation/Testing.md) methods.
  
## Checklist


Workbooks/CosmosDb/Throttling Insights/Throttling Insights.workbook
```diff
@@ -0,0 +1,1441 @@
...
+                    "metrics": [
+                      {
+                        "namespace": "microsoft.documentdb/databaseaccounts",
+                        "metric": "microsoft.documentdb/databaseaccounts-Requests-NormalizedRUConsumption",
+                        "aggregation": 3,
+                        "splitBy": [
+                          "PartitionKeyRangeId"
+                        ],
+                        "splitBySortOrder": -1,
+                        "splitByLimit": 1000
```
StefArroyo: Could we validate this chart with a large account/container that has several thousand partition key ranges? `splitByLimit` caps the chart at 1,000 series, so I would like to understand what the customer sees beyond that scale: which 1,000 ranges are selected, whether the hottest/throttled range is guaranteed to remain visible, and whether the workbook clearly indicates that results were truncated. A screenshot or test using a large account would help. If the chart silently truncates, could we show a top-N hottest view plus an explicit "showing 1,000 of N ranges" message?

iriaosara: Yes. Made the changes. Have the screen recording here https://microsoft-my.sharepoint.com/:f:/p/iriaosara/IgAofQLQy1CXSaSD5s0ihszNATsiIc3TcqZnzE5bzEOIZ_A?e=roaZ8a

StefArroyo (approved):

